PR #26, opened 2018-03-12
Add ability to setup environment variables before calling Maven

jeffgbutler:
Currently, the maven integration only works if the JAVA_HOME environment variable is set permanently.  But setting JAVA_HOME permanently is inconvenient in many cases - especially if we are testing against many different versions of Java.

There is an existing configuration property in the VS Code Java support to configure JAVA_HOME - the "java.home" setting.

This PR adds a configuration property to the maven extension: "maven.set.javaHome".  If true, and if java.home is set, then the JAVA_HOME environment variable will be set to the configured value when a new terminal is opened.

I've tested this with Powershell, CMD, and GitBash and it works as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,10 +96,10 @@ This extension (Maven for Java) can reuse that setting if you desire:
     }
 ```
 
-With this support, you can specify JAVA_HOME in one place and you don't need to use the `maven.terminal.customEnv` setting unless
+With this support, you can specify JAVA_HOME in one place and you do not need to use the `maven.terminal.customEnv` setting unless
 you have other environment variables to set.
 
-If you have JAVA_HOME configured through the `maven.terminal.customEnv` setting, and also specify to reuse the Red Hat setting, then value from `maven.terminal.customEnv` will take precedent.
+If you have JAVA_HOME configured through the `maven.terminal.customEnv` setting, and also specify to reuse the Red Hat setting, then the value from `maven.terminal.customEnv` will take precedent.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ This extension (Maven for Java) can reuse that setting if you desire:
 With this support, you can specify JAVA_HOME in one place and you do not need to use the `maven.terminal.customEnv` setting unless
 you have other environment variables to set.
 
-If you have JAVA_HOME configured through the `maven.terminal.customEnv` setting, and also specify to reuse the Red Hat setting, then the value from `maven.terminal.customEnv` will take precedent.
+If you have JAVA_HOME configured through the `maven.terminal.customEnv` setting, and also specify to reuse the Red Hat setting, then the value from `maven.terminal.customEnv` will take precedence.
 
 ## Contributing
 
--- a/README.md
+++ b/README.md
@@ -58,6 +58,49 @@ Provide Maven executable filepath.
 
 VS Code collects usage data and sends it to Microsoft to help improve our products and services. Read our [privacy statement](http://go.microsoft.com/fwlink/?LinkId=521839) to learn more. If you don’t wish to send usage data to Microsoft, you can set the `telemetry.enableTelemetry` setting to `false`. Learn more in our [FAQ](https://code.visualstudio.com/docs/supporting/faq#_how-to-disable-telemetry-reporting).
 
+## JAVA_HOME and Other Environment Variables
+
+This extension executes Maven by opening a terminal session and then calling Maven in that session.
+Maven requires the JAVA_HOME environment variable to be set. Maven will also look for other variables such as MAVEN_OPTS. If you prefer not to set those variables permanently you can configure them, or any other environment variable, in settings:
+
+```
+    {
+        "maven.terminal.customEnv": [
+            {
+                "environmentVariable": "MAVEN_OPTS",               // variable name
+                "value": "-Xms1024m -Xmx4096m"                     // value
+            },
+            {
+                "environmentVariable": "JAVA_HOME",                // variable name
+                "value": "C:\\Program Files\\Java\\jdk-9.0.4"      // value
+            }
+        ]
+    }
+```
+
+### Special Handling for JAVA_HOME
+
+If you have Red Hat's Java Language Support extension installed, then you can specify JAVA_HOME in settings for that extension:
+
+```
+    {
+        "java.home": "C:\\Program Files\\Java\\jdk-9.0.4"      // Red Hat Java Language Support Setting
+    }
+```
+
+This extension (Maven for Java) can reuse that setting if you desire:
+
+```
+    {
+        "maven.terminal.useJavaHome": true      // Use the Red Hat Java Language Support Setting for JAVA_HOME
+    }
+```
+
+With this support, you can specify JAVA_HOME in one place and you don't need to use the `maven.terminal.customEnv` setting unless
+you have other environment variables to set.
+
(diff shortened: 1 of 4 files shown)


src/VSCodeUI.ts
```diff
@@ -53,6 +54,32 @@ export namespace VSCodeUI {
         }
     }
 
+    export function setJavaHomeIfRequired(terminal: Terminal): void {
+        const javaHome: string = workspace.getConfiguration("java").get<string>("home");
```
Eskibear: this extension does not depend on `redhat.java`, it seems strange to read configuration like `java.home` from that extension. 😕

src/VSCodeUI.ts
```diff
@@ -53,6 +54,32 @@ export namespace VSCodeUI {
...
+        const setJavaHome: boolean = workspace.getConfiguration("maven").get<boolean>("set.javaHome");
+        if (setJavaHome && javaHome) {
```
Eskibear: What if the user needs to specify `JAVA_HOME` by his/her own without installing the redhat java extension?
So is it good to allow users to do that by 2 configurations as following: 
```
          "maven.javaHome.from": {
            "type":"string",
            "default": "system envrionment",
            "enum": [
              "system envrionment",
              "extension: redhat.java",
              "specified in maven.javaHome.path"
            ]
          },
          "maven.javaHome.path": {
            "type" :"string",
            "default": ""
          }
```
Or is it too complicated/noisy to have so many configuration entries?

jdneo:
I prefer the `maven.terminal.customEnv` solution. Adding the `maven.javaHome.from` looks too complicated to me

jeffgbutler:
I think you have to consider for most users the Redhat extension will be installed because of the bundled Java extension pack.  If we only do the `maven.terminal.customEnv` solution, then a user would need to configure the Java home value twice - and that seems kind of silly to me.

I think there needs to be a way to use `java.home` if it is there.  But I recognize there is an edge case here too.  Maybe there  should be two properties.  `maven.terminal.customEnv` for the edge case and a Boolean `maven.use.java.home` for most users.

package.json
```diff
@@ -293,6 +293,32 @@
             "default": "",
             "description": "Specifies absolute path of mvn executable.",
             "scope": "window"
+          },
+          "maven.terminal.use.java.home": {
```
Eskibear: `use.java.home` seems too long or too deep in level? I prefer `useJavaHome`.

package.json
```diff
@@ -293,6 +293,32 @@
...
+          },
+          "maven.terminal.customEnv": {
+            "type": "array",
+            "items": {
+              "type": "object",
+              "title": "environment setting",
+              "properties": {
+                "environmentVariable": {
+                  "type": "string",
+                  "description": "environment variable to set"
```
Eskibear: Should be more human-readable. e.g. "Name of the environment variable to set."

package.json
```diff
@@ -293,6 +293,32 @@
...
+                },
+                "value" : {
+                  "type":"string",
+                  "description": "value for the variable"
```
Eskibear: Should be more human-readable.

package.json
```diff
@@ -293,6 +293,32 @@
...
+                }
+              }              
+            },
+            "default": [],
+            "description": "An array of environment settings.  These will be added to the terminal session before maven is first run",
```
Eskibear: Please improve the wording. E.g. there are 2 spaces before the word "These" now, and missing period in the end.

src/VSCodeUI.ts
```diff
@@ -53,6 +54,47 @@ export namespace VSCodeUI {
...
+
+        const environmentSettings: EnvironmentSetting[] = workspace.getConfiguration("maven").get("terminal.customEnv");
+        environmentSettings.forEach((s: EnvironmentSetting) => {
+            terminal.sendText(getEnvironmentVariableCommand(s.environmentVariable, s.value), true);
+        });
+    }
+
+    export function setJavaHomeIfAvailable(terminal: Terminal): void {
+        const javaHome: string = workspace.getConfiguration("java").get<string>("home");
+        const useJavaHome: boolean = workspace.getConfiguration("maven").get<boolean>("terminal.use.java.home");
```
Eskibear: if the configuration name `terminal.use.java.home` is changed, don't forget to update here.

src/VSCodeUI.ts
```diff
@@ -53,6 +54,47 @@ export namespace VSCodeUI {
...
+        };
+
+        const environmentSettings: EnvironmentSetting[] = workspace.getConfiguration("maven").get("terminal.customEnv");
+        environmentSettings.forEach((s: EnvironmentSetting) => {
+            terminal.sendText(getEnvironmentVariableCommand(s.environmentVariable, s.value), true);
+        });
+    }
+
+    export function setJavaHomeIfAvailable(terminal: Terminal): void {
+        const javaHome: string = workspace.getConfiguration("java").get<string>("home");
```
Eskibear: suggest to add comments indicating that the conf `java.home` is from extension `redhat.java`, so as not to confuse future developers.

src/VSCodeUI.ts
```diff
@@ -53,6 +54,47 @@ export namespace VSCodeUI {
...
+        const useJavaHome: boolean = workspace.getConfiguration("maven").get<boolean>("terminal.use.java.home");
+        if (useJavaHome && javaHome) {
+            terminal.sendText(getEnvironmentVariableCommand("JAVA_HOME", javaHome), true);
+        }
+    }
+
+    export function getEnvironmentVariableCommand(variable: string, value: string): string {
```
Eskibear: confusing name, might be `setEnvironmentVariableCommand`? Is it a typo?

src/VSCodeUI.ts
```diff
@@ -53,6 +54,47 @@ export namespace VSCodeUI {
         }
     }
 
+    export function setupEnvironment(terminal: Terminal): void {
+        // do this first so it can be overridden if desired
+        setJavaHomeIfAvailable(terminal);
```
Eskibear: According to the call sequence of `setJavaHomeIfAvailable` and `environmentSettings.forEach`, if I set both `java.home` and `maven.terminal.customEnv[JAVA_HOME]`, the latter overrides the former. Is it the expected behaviour for most users? Or how to avoid ambiguity?

jeffgbutler: IMHO, that is expected behavior.  If `java.home` is configured, then we can reuse it.  But if a user also configures `JAVA_HOME` for our extension, then it should take precedence.

I will explain that in the README.

src/VSCodeUI.ts
jeffgbutler: It is a "get" that returns a "set" command string (ha ha).  I'll figure out a different name.

jeffgbutler:
I think I have made all your requested changes.  Hopefully ready to merge 😃

README.md
```diff
@@ -58,6 +58,49 @@ Provide Maven executable filepath.
...
+```
+    {
+        "maven.terminal.useJavaHome": true      // Use the Red Hat Java Language Support Setting for JAVA_HOME
+    }
+```
+
+With this support, you can specify JAVA_HOME in one place and you do not need to use the `maven.terminal.customEnv` setting unless
+you have other environment variables to set.
+
+If you have JAVA_HOME configured through the `maven.terminal.customEnv` setting, and also specify to reuse the Red Hat setting, then the value from `maven.terminal.customEnv` will take precedent.
```
Eskibear: take precedence ?

jeffgbutler:
Done

PR #83, opened 2018-07-16
Refactor maven projects explorer

Eskibear:
Original PR #81. This one has a clean commit history.


src/Utils.ts
```diff
@@ -216,7 +211,7 @@ export namespace Utils {
         });
     }
 
-    export async function findAllInDir(currentPath: string, targetFileName: string, depth: number, exclusion: string[] = ["**/.*"]): Promise<string[]> {
+    async function findAllInDir(currentPath: string, targetFileName: string, depth: number, exclusion?: string[]): Promise<string[]> {
```
jdneo: Is it possible to use `workspace.findFiles()`

src/Utils.ts
```diff
@@ -249,19 +244,66 @@ export namespace Utils {
...
     }
 
-    export function executeMavenCommand(command: string, pomfile: string): void {
+    function getMaven(): string {
+        const executablePathInConf: string = workspace.getConfiguration("maven.executable").get<string>("path");
+        // TODO: use maven wrapper if path is not set and mvnw is found.
+        return executablePathInConf ? executablePathInConf : "mvn";
+    }
+
+    function wrapMaven(mvn: string): string {
```
jdneo: wrapMaven -> wrapMavenWithQuote?

src/Utils.ts
```diff
@@ -249,19 +244,66 @@ export namespace Utils {
...
+    export async function executeInBackground(command: string, pomfile: string): Promise<{}> {
+        const mvnString: string = wrapMaven(getMaven());
+
+        const fullCommand: string = [
+            mvnString,
+            command.trim(),
+            "-f",
+            `"${formattedFilepath(pomfile)}"`,
+            workspace.getConfiguration("maven.executable", Uri.file(pomfile)).get<string>("options")
+        ].filter((x: string) => x).join(" ");
```
jdneo: Boolean?

src/Utils.ts
```diff
@@ -249,19 +244,66 @@ export namespace Utils {
...
+
+        const rootfolder: WorkspaceFolder = workspace.getWorkspaceFolder(Uri.file(pomfile));
+        const customEnv: {} = VSCodeUI.setupEnvironment();
+        const execOptions: child_process.ExecOptions = {
+            cwd: rootfolder ? rootfolder.uri.fsPath : path.dirname(pomfile),
```
jdneo: Why we need path.dirname(pomfile) here?

Eskibear: it was to fix mvnw issue months ago. Actually it should be `path.dirname(path of mvnw)`. Will force to use maven wrapper if found.

PR #96, opened 2018-07-19
avoid to show duplicate archetypes

### LINKED ISSUE #93: 1st page of project generation: only show archetypes in local catalog but not in remote catalog

Eskibear:
fix #93

### DIFF
--- a/src/archetype/Archetype.ts
+++ b/src/archetype/Archetype.ts
@@ -11,10 +11,10 @@ export class Archetype {
         return `${this.groupId}:${this.artifactId}`;
     }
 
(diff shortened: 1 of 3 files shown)


src/archetype/Archetype.ts
```diff
@@ -7,11 +7,14 @@ export class Archetype {
     public repository: string;
     public description: string;
     public versions: string[];
+    public get identifier(): string {
+        return `${this.groupId}:${this.artifactId}`;
+    }
 
-    constructor(aid: string, gid: string, repo?: string, desc?: string) {
+    constructor(aid: string, gid: string, repo?: string, desc?: string, versions?: string[]) {
```
jdneo: argument with default value instead of optional argument

PR #104, opened 2018-07-20
fix warning: explicitly specify null for resource scope conf

Eskibear:
To fix warining message

Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for 'maven.executable.path', provide the URI of a resource or 'null' for any resource.

### DIFF
--- a/src/Settings.ts
+++ b/src/Settings.ts
@@ -14,7 +14,7 @@ export namespace Settings {
         }
     }
 
-    export function excludedFolders(resource?: Uri): string[] {
+    export function excludedFolders(resource: Uri): string[] {
         return _getMavenSection("excludedFolders", resource);
     }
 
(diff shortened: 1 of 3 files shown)


src/Settings.ts
```diff
@@ -40,6 +40,6 @@ export namespace Settings {
     }
 
     function _getMavenSection<T>(section: string, resource?: Uri): T {
-        return workspace.getConfiguration("maven", resource).get<T>(section);
+        return workspace.getConfiguration("maven", resource || null).get<T>(section);
```
jdneo: Make the default value of resource to `null` instead of making it as an optional parameter?

Eskibear: `export function getConfiguration(section?: string, resource?: Uri)` 
 if resource is `undefined`, it gets warning. It suggests you to explicitly pass a `null` if you need to get conf for any resource.

Eskibear: And if you think this is wired, we can keep it as optional parameter, and ensure to pass a `null` when calling it.

PR #118, opened 2018-09-01
Sort maven project list in Explorer and hide invalid pom.xml files

owenconti:
- Fixed a TS typing error on a boolean variable.
- Updated `_ensureMavenProjectParsed` of `MavenProjectNode` to become public.
- Alphabetically sort the array of `MavenProjectNode` items in `WorkspaceFolderNode`.

screen shot 2018-08-31 at 6 21 31 pm

If I've made any mistakes or if there is a better way of approaching the problem, please let me know.

### DIFF
--- a/src/explorer/model/MavenProjectNode.ts
+++ b/src/explorer/model/MavenProjectNode.ts
@@ -18,7 +18,7 @@ export class MavenProjectNode extends NodeBase {
     }
 
     public async getTreeItem(): Promise<vscode.TreeItem> {
-        await this._ensureMavenProjectParsed();
+        await this.ensureMavenProjectParsed();
         const treeItem: vscode.TreeItem = new vscode.TreeItem(this._mavenProject.name);
         treeItem.iconPath = {
             light: Utils.getResourcePath("project.svg"),
@@ -53,7 +53,7 @@ export class MavenProjectNode extends NodeBase {
         return this._mavenProject.modules.length > 0;
     }
 
-    private async _ensureMavenProjectParsed(): Promise<void> {
+    public async ensureMavenProjectParsed(): Promise<void> {
         if (!this._mavenProject) {
             const pom: {} = await Utils.parseXmlFile(this._pomPath);
             this._mavenProject = new MavenProject(pom);
--- a/src/explorer/model/MavenProjectNode.ts
+++ b/src/explorer/model/MavenProjectNode.ts
@@ -19,6 +19,7 @@ export class MavenProjectNode extends NodeBase {
 
     public async getTreeItem(): Promise<vscode.TreeItem> {
         await this.ensureMavenProjectParsed();
+
         const treeItem: vscode.TreeItem = new vscode.TreeItem(this._mavenProject.name);
         treeItem.iconPath = {
             light: Utils.getResourcePath("project.svg"),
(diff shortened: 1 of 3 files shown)


owenconti:
Hi @Eskibear,

I've updated the code to handle errors thrown in the `ensureMavenProjectParsed` function. I also updated `getChildren` in `WorkspaceFolderNode` to make sure the `MavenProjectNode` is valid before considering it a children. 

Here is a screenshot showing the difference:

screen shot 2018-09-01 at 4 37 06 pm

Valid pom.xml ^

screen shot 2018-09-01 at 4 37 16 pm

Invalid pom.xml ^ (notice the project missing from the Explorer list)

src/explorer/model/MavenProjectNode.ts
```diff
@@ -49,14 +50,24 @@ export class MavenProjectNode extends NodeBase {
...
+        await this.ensureMavenProjectParsed();
+        return !!this._mavenProject;
+    }
+
     private get _hasModules(): boolean {
         return this._mavenProject.modules.length > 0;
     }
 
-    private async _ensureMavenProjectParsed(): Promise<void> {
+    public async ensureMavenProjectParsed(): Promise<void> {
```
Eskibear: `ensureMavenProjectParsed` actually doesn't ensure a maven project is successfully parsed, `isValidPom` should always be checked after calling this function. So I think it's better to make this private, and only to expose `isValidPom`.

src/explorer/model/MavenProjectNode.ts
```diff
@@ -49,14 +50,24 @@ export class MavenProjectNode extends NodeBase {
     public get mavenProject(): MavenProject {
         return this._mavenProject;
     }
+
+    public async isValidPom(): Promise<boolean> {
```
Eskibear: the class is `MavenProjectNode`, so `hasValidPom` might be a better name.

src/explorer/model/MavenProjectNode.ts
```diff
@@ -18,7 +18,8 @@ export class MavenProjectNode extends NodeBase {
     }
 
     public async getTreeItem(): Promise<vscode.TreeItem> {
-        await this._ensureMavenProjectParsed();
+        await this.ensureMavenProjectParsed();
```
Eskibear: There is a potential NPE for `this._mavenProject.name` in next line. Although currently in this PR projects with invalid pom have been filtered out in `WorkspaceFolder.getChildren`, it's safer to check it here to avoid potential NPEs in future.

e.g. something like
```typescript
if (! await this.isValidPom()) {
    return undefined;
} 
```

src/explorer/model/WorkspaceFolderNode.ts
```diff
@@ -39,4 +49,10 @@ export class WorkspaceFolderNode extends NodeBase {
     private async _searchForPomPaths(): Promise<void> {
         this._pomPaths = await Utils.getAllPomPaths(this._workspaceFolder);
     }
+
+    private async _sortChildren(): Promise<void> {
```
Eskibear: I don't see any async things inside, so I think there's no need to declare as an `async` function.

src/explorer/model/WorkspaceFolderNode.ts
```diff
@@ -28,7 +28,17 @@ export class WorkspaceFolderNode extends NodeBase {
...
+
+        for ( const pomPath of this._pomPaths ) {
+            const projectNode: MavenProjectNode = new MavenProjectNode(pomPath);
+
+            if ( await projectNode.isValidPom() ) {
+                this._children.push(projectNode);
+            }
+        }
+
+        await this._sortChildren();
```
Eskibear: no need to use async/await for the _sortChildren function.

owenconti:
I think I got all the changes. I also changed `_ensureMavenProjectParsed` to `_parseMavenProject` to more accurately describe what it does.

Apologies, I hadn't seen that you made an issue for the parsing of invalid pom.xml files.

PR #143, opened 2018-11-09
Better support for WSL

### LINKED ISSUE #142: maven goals fail to execute when using WSL bash for Terminal
**Describe the bug**
maven goals fail to execute when using WSL bash for Terminal

**To Reproduce**
1. open a maven based project on windows
2. Have WSL install  (I am using Ubuntu)
3. Set VSCode to use bash.exe
  - "terminal.integrated.shell.windows": "C:\\Windows\\System32\\bash.exe"
  - based on this https://code.visualstudio.com/docs/editor/integrated-terminal#_windows
4. Attempt to use a maven goal by right-clicking on a goal "validate"

**Expected behavior**
To run
mvn validate -f ${project_path}/pom.xml

**Environments (please complete the following information as much as possible):**
* Version: 1.28.2 (user setup)
* Commit: 7f3ce96ff4729c91352ae6def877e59c561f4850
* Date: 2018-10-17T00:23:51.859Z
* Electron: 2.0.9
* Chrome: 61.0.3163.100
* Node.js: 8.9.3
* V8: 6.1.534.41
* Architecture: x64
* Maven for Java: 0.11.1

**Screenshots**
maven_for_maven_wsl_broken

**Additional context**
It seems the code to detect that WSL is being used is very particular, and thus fails to correctly Identify the environment.

RobertDeRose:
* Changed `currentWindowsShell` to be more generic and straighforward
* Changed `toWSLPath` to return path converting `\` to `/` even when
  not containing a `:\\`

closes #142

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -265,7 +265,7 @@ export namespace Utils {
             return 'Command Prompt';
         } else if (currentWindowsShellPath.endsWith("powershell.exe")) {
             return 'PowerShell';
-        } else if (currentWindowsShellPath.endsWith("bash.exe")) {
+        } else if (currentWindowsShellPath.endsWith("bash.exe") || currentWindowsShellPath.endsWith("wsl.exe")) {
             if (currentWindowsShellPath.includes("Git")) {
                 return 'Git Bash';
             }
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -260,27 +260,19 @@ export namespace Utils {
     }
 
     export function currentWindowsShell(): string {
-        const is32ProcessOn64Windows: boolean = process.env.hasOwnProperty('PROCESSOR_ARCHITEW6432');
-        const system32Path: string = `${process.env.windir}\\${is32ProcessOn64Windows ? 'Sysnative' : 'System32'}`;
-        const expectedLocations: { [shell: string]: string[] } = {
-            'Command Prompt': [`${system32Path}\\cmd.exe`],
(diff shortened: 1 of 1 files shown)


src/Utils.ts
```diff
@@ -260,27 +260,19 @@ export namespace Utils {
...
-                return key;
+        if (currentWindowsShellPath.endsWith("cmd.exe")) {
+            return 'Command Prompt';
+        } else if (currentWindowsShellPath.endsWith("powershell.exe")) {
+            return 'PowerShell';
+        } else if (currentWindowsShellPath.endsWith("bash.exe")) {
+            if (currentWindowsShellPath.includes("Git")) {
+                return 'Git Bash';
             }
+            return 'WSL Bash';
```
Eskibear: Previous logic was following VSCode. Your change looks good, and now WSL can be also launched by a wrapper named `wsl.exe`, and I think it's better to support it.

Refer to [these lines](https://github.com/Microsoft/vscode/blob/758206be09ddfaac46e941a77375ec34ed928b4b/src/vs/workbench/parts/terminal/electron-browser/terminalService.ts#L218-L238).

RobertDeRose: I added support for `wsl.exe`. I didn't know it even existed, but it's cool because in Ubuntu WSL I set zsh.exe as my shell, and wsl.exe starts zsh.exe without the need for loading bash.

RobertDeRose:
Thanks for the merge!

PR #154, opened 2018-11-27
Use vscode@1.1.22 to fix vulnerability issue of event-stream.

Eskibear:
To resolve the alert. 
```
Vulnerable versions: > 3.3.4
Patched version: No fix
The NPM package flatmap-stream is considered malicious. A malicious actor added this package as a dependency to the NPM event-stream package in versions 3.3.6 and later. Users of event-stream are encouraged to downgrade to the last non-malicious version, 3.3.4.

Users of flatmap-stream are encouraged to remove the dependency entirely.
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -363,7 +363,7 @@
     "tslint": "^5.8.0",
     "tslint-microsoft-contrib": "^5.0.1",
     "typescript": "^2.9.2",
-    "vscode": "^1.1.21"
+    "vscode": "^1.1.22"
   },
   "dependencies": {
     "fs-extra": "^4.0.3",
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,15 +59,15 @@
             }
         },
         "ajv": {
-            "version": "5.5.2",
-            "resolved": "https://registry.npmjs.org/ajv/-/ajv-5.5.2.tgz",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -363,7 +363,7 @@
     "tslint": "^5.8.0",
     "tslint-microsoft-contrib": "^5.0.1",
     "typescript": "^2.9.2",
-    "vscode": "^1.1.21"
+    "vscode": "^1.1.22"
```
testforstephen: why need to update vscode?

Eskibear: the dependency is introduced by vscode.
```bash
$> npm ls event-stream
vscode-maven@0.11.2 /Users/sechs/Documents/vscode-maven
└─┬ vscode@1.1.21
  ├─┬ gulp-remote-src-vscode@0.5.0
  │ └── event-stream@3.3.5
  ├─┬ gulp-symdest@1.1.0
  │ └── event-stream@3.3.5  deduped
  ├─┬ gulp-untar@0.0.7
  │ └── event-stream@3.3.5  deduped
  └─┬ gulp-vinyl-zip@2.1.0
    └── event-stream@3.3.5  deduped

```

testforstephen: Is event-stream an implicit dependency of vscode package?

PR #158, opened 2018-12-05
Allow to manually refresh plugin goals.

Eskibear:
image

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -8,6 +8,7 @@ import { dispose as disposeTelemetryWrapper, initializeFromJsonFile, instrumentO
 import { ArchetypeModule } from "./archetype/ArchetypeModule";
 import { OperationCanceledError } from "./Errors";
 import { mavenExplorerProvider } from "./explorer/MavenExplorerProvider";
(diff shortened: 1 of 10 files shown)


src/extension.ts
```diff
@@ -65,8 +66,8 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
         });
     });
 
-    registerCommand(context, "maven.project.refreshAll", (): void => {
-        mavenExplorerProvider.refresh();
+    registerCommand(context, "maven.project.refreshAll", (item?: ITreeItem): void => {
```
Eskibear: I've reused the command `refreshAll`, so it's now a "general" refresh action. Do you think I should change the command name? @jdneo

Eskibear:
By the above commit 6a85b11, the project items under `Modules` are exactly the same instances under the root. So it looks more like a link now, and it avoids to calculate effective-pom again.

src/extension.ts
jdneo: I prefer `maven.explorer.refresh`

PR #159, opened 2018-12-05
Better UX for Effective POM. 

Eskibear:
1. Read calculated Effective-POM if possible. 
2. Add listeners to update Effective-POM automatically.

### DIFF
--- a/src/Utils.ts
+++ b/src/Utils.ts
@@ -313,22 +313,15 @@ export namespace Utils {
     }
 
     export async function showEffectivePom(pomPath: string): Promise<void> {
-        const pomxml: string = await window.withProgress({ location: ProgressLocation.Window }, (p: Progress<{ message?: string }>) => new Promise<string>(
-            async (resolve, reject): Promise<void> => {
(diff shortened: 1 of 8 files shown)


src/Utils.ts
```diff
@@ -313,34 +313,37 @@ export namespace Utils {
...
+        if (project) {
+            await project.calculateEffectivePom();
+            pomxml = project.rawEffectivePom;
+        } else {
+            pomxml = await Utils.getEffectivePom(pomPath);
+        }
+
+        if (pomxml) {
+            const document: TextDocument = await workspace.openTextDocument({ language: "xml", content: pomxml });
+            window.showTextDocument(document);
```
jdneo: Show in the active ViewColumn?

Eskibear: Sounds a good idea. So I suppose by using ViewColumn, I can specify the title of it, e.g. `[Project 1]Effective-POM`?

Eskibear: Got it. Updated.

PR #169, opened 2018-12-10
Use globalState to cache plugin info.

Eskibear:
Use a `PluginInfoProvider` to get detailed info of a maven plugin, e.g. prefix, goals.

It uses the `ExtensionContext.globalState` as the cache to improve perf.

### DIFF
--- a/src/explorer/PluginInfoProvider.ts
+++ b/src/explorer/PluginInfoProvider.ts
@@ -1,0 +1,89 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as _ from "lodash";
+import * as vscode from "vscode";
+import { Utils } from "../Utils";
+
+const KEY_PLUGINS: string = "plugins";
+
+interface IPluginInfo {
+    prefix?: string;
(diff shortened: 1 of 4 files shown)


src/explorer/PluginInfoProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+            const goalsMatch: string[] = rawOutput.match(goalRegExp);
+            if (goalsMatch) {
+                goals.push(...goalsMatch);
+            }
+        }
+
+        return { prefix, goals };
+    }
+
+    private pluginId(gid: string, aid: string, version?: string): string {
```
jdneo: Add a verb to the method name.

src/explorer/PluginInfoProvider.ts
```diff
@@ -0,0 +1,89 @@
...
+        return `${gid}:${aid}${version ? `:${version}` : ""}`;
+    }
+}
+
+// tslint:disable-next-line:export-name
```
jdneo: Rename the file to `pluginInfoProvider.ts` instead of disabling the tslint rule.

PR #177, opened 2018-12-13
Refactor: UI Utils.

Eskibear:
Nothing new, just better organization.

### DIFF
--- a/src/mavenOutputChannel.ts
+++ b/src/mavenOutputChannel.ts
@@ -1,0 +1,31 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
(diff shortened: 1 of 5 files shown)


src/mavenOutputChannel.ts
```diff
@@ -0,0 +1,31 @@
...
+
+    public append(message: any): void {
+        this.channel.append(message);
+    }
+
+    public show(): void {
+        this.channel.show();
+    }
+
+    public dispose(): void {
```
jdneo: Call this method in deactivate().

Eskibear: Good catch. Will update.

Eskibear: done.

PR #180, opened 2018-12-14
Improve performance of Maven explorer

Eskibear:
1. Now effective-pom is not calculated until the user expands the "Plugins" node. 
2. When the "Plugins" node is expanded, plugin nodes are rendered with the qualified names. A corresponding mvn command to fetch plugin info is queued to execute for each plugin. After acquiring the info (prefix), it refreshes the corresponding node, putting the prefix ahead.


src/taskExecutor.ts
```diff
@@ -0,0 +1,40 @@
+class Queue<T> {
```
jdneo: license

src/taskExecutor.ts
```diff
@@ -0,0 +1,40 @@
...
+        this._store.push(val);
+    }
+    public pop(): T | undefined {
+        return this._store.shift();
+    }
+}
+
+class TaskExecutor {
+    private _tasks: Queue<any> = new Queue();
+    private _executing: boolean = false;
```
jdneo: isExecuting

src/taskExecutor.ts
```diff
@@ -0,0 +1,40 @@
...
+    public push(val: T): void {
+        this._store.push(val);
+    }
+    public pop(): T | undefined {
+        return this._store.shift();
+    }
+}
+
+class TaskExecutor {
+    private _tasks: Queue<any> = new Queue();
```
jdneo: Add `dispose()` function which clear the queue when deactivating

src/taskExecutor.ts
Eskibear: added

src/taskExecutor.ts
Eskibear: added.

Issue #195, opened 2018-12-27
Auto complete for dependencies for pom.xml


Eskibear:
The upstream issue is #47 

An important feature to improve pom.xml editing experience is **dependency auto-completion**. 
The basic use cases are:
1. For `<groupId>[cursor]</groupId>`, show a list of available group ids, e.g. from local maven repository.
2. For `<artifactId>[cursor]</artifactId>`, show a list of available artifact ids according to the value of groupId.

Eskibear:
### Experiences in Java IDEs

#### Eclipse photon
I don't see any completion for groupId or artifactId. 

#### IntelliJ IDEA 2018.3
* For groupId, it always provides a list of all group ids which contains the keyword you've typed in. 
* For artifactId, it lists artifacts of the given group id.
* For version, it lists versions in the same way above, together with two extra items ("RELEASE" and "LATEST")

Note that all the completion items are from local Maven repository. Every time you open a Maven project, it takes some time to update the Maven indices of your local repository. If you don't have an artifact installed into your local repository, it won't appear in the completion list.

Eskibear:
We need context info of current active pom.xml. E.g. we need to know current cursor position, and if it's located in an `artifactId` node, the value of `groupId` is required to calculate available artifacts. And if it's located elsewhere, we don't need to provide the result.

1. `provideCompletionItems` has the cursor position as a param, so it's easy to get the whole text before the cursor.
2. `xml-zero-lexer` is a lexer which can tokenize the xml file into an array of tokens, so it's feasible to tell which node the cursor is in.

Eskibear:
The current implementation requires to find all pom files in local Maven Repo, which introduces extra CPU & IO overload. It depends on the size of the local Maven repo, and I'm not sure whether it will lead to performance deterioration. So I think we can disable it by default, and add a flag in settings to enable the feature.

Eskibear:
With PR #221 , it's able to consume data from Maven Central repository. 
With PR #222 , it improves performance of fetching data from Maven local repository.

TODO
- [ ] Refine sorting of completion items, e.g. versions should be sorted semantically
- [ ] Use effective local repository instead of the hard-coded `~/.m2/repository`

PR #196, opened 2018-12-27
Provide completion items only for dependency nodes.

Eskibear:
See #195 

1. Use `xml-zero-lexer`  to tokenize pom.xml
2. Provider completion results only when cursor is in node `dependency>groupId` or `dependency>artifactId`

### DIFF
--- a/src/completionProvider.ts
+++ b/src/completionProvider.ts
@@ -2,11 +2,102 @@
 // Licensed under the MIT license.
 
 import * as vscode from "vscode";
+import Lexx from "xml-zero-lexer";
 
 class CompletionProvider implements vscode.CompletionItemProvider {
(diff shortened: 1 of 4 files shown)


src/completionProvider.ts
```diff
@@ -2,11 +2,102 @@
...
+    public addChild(child: ElementNode): void {
+        if (this.value !== undefined) {
+            this.value = undefined; // value cannot exist with children.
+        }
+        if (this.children === undefined) {
+            this.children = [];
+        }
+        this.children.push(child);
+    }
+
```
jdneo: Remove the empty line next PR

Eskibear: ok

PR #197, opened 2018-12-27
Complete groupId & artirfactId from local Maven repo.

Eskibear:
Part of workitem #195 

1. Scan local maven repository on activation. 
2. Complete groupId & artifactId with the data.


package.json
```diff
@@ -400,7 +400,10 @@
     "webpack-cli": "^3.1.0"
   },
   "dependencies": {
+    "@types/glob": "^7.1.1",
```
jdneo: Move to dev-dependency

src/completionProvider.ts
```diff
@@ -1,11 +1,58 @@
...
+import * as path from "path";
 import * as vscode from "vscode";
 import Lexx from "xml-zero-lexer";
 
+type Artifact = {
+    groupId: string;
+    artifactId: string;
+    version: string;
+    pom: string;
+};
```
jdneo: Add an empty line here

Eskibear: will remove `Artifact` because we only need to save a list of versions here.

PR #222, opened 2019-01-23
Retrieve completion items from local Maven repository on the fly

Eskibear:
Original issue: #195 , Continuing work of #197 

### In this PR
- Remove indexing work which cause huge time and CPU cost.
- Searching local Maven repository on completion, with depth of 2. (good performance)
- Typo fixing

### Todo next
- Stop using hard-coded m2_home
- Sort versions in reverse order.

### DIFF
--- a/package.json
+++ b/package.json
@@ -398,12 +398,6 @@
             "description": "%configuration.maven.terminal.customEnv%",
             "scope": "window"
           },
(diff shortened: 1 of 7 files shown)


Eskibear:
The behavior is: 
maven-completion

PR #229, opened 2019-01-24
Remove duplicated items in completion list

Eskibear:
Previously it uses two completion providers, with duplicated results sometimes.

Now combine them into one, removing the dup items.

### DIFF
--- a/src/completion/IArtifactProvider.ts
+++ b/src/completion/IArtifactProvider.ts
@@ -3,7 +3,7 @@
 
 import { CompletionItem } from "vscode";
 
-export interface IArtifactProvider {
+export interface IMavenCompletionItemProvider {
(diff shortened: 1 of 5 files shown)


Eskibear:
it's part of the work item #195

src/completion/IArtifactProvider.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import { CompletionItem } from "vscode";
+
+export interface IArtifactProvider {
```
jdneo: Rename. Suggestion: IMavenCompletionItemProvider

Eskibear: done.

PR #230, opened 2019-01-24
Update groupId together if possible when completing artifactId

Eskibear:
This PR adds additionalTextEdit for artifact ID completion items to achieve the goal.

1. Changed `ElementNode` as we need to know both start/end position of groupId node. 
2. Information of group Id is carried in CompletionItem.data (standard LSP, but not defined in vscode.d.ts)

### DIFF
--- a/src/completion/lexerUtils.ts
+++ b/src/completion/lexerUtils.ts
@@ -72,7 +72,7 @@ function getElementHierarchy(text: string, tokens: number[][], cursorOffset: num
         switch (token[0]) {
             case NodeTypes.XML_DECLARATION:
             case NodeTypes.ELEMENT_NODE: {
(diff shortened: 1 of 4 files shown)


src/completion/lexerUtils.ts
```diff
@@ -60,41 +61,101 @@ export function getCurrentNode(document: vscode.TextDocument, position: vscode.P
...
+                if (currentNode) {
+                    currentNode.addChild(newElement);
                 }
-                iter = newElement;
+
+                pointer = end + 1; // pass ">" mark.
+                elementNodes.push(newElement);
+                newElement.contentStart = pointer;
+                break;
+
```
jdneo: remove empty line

PR #235, opened 2019-01-28
send telemetry when completion item selected

Eskibear:
- It calls command `maven.completion.selected` whenever an item is selected.
- It sends `{completeFor: string, source: string}` as the info. 
  - completeFor: dependency/plugin/groupId/artifactId/version
  - source: snippet/maven-local/maven-central

### DIFF
--- a/src/completion/centralProvider.ts
+++ b/src/completion/centralProvider.ts
@@ -3,11 +3,10 @@
 
 import * as _ from "lodash";
 import * as vscode from "vscode";
+import { COMMAND_COMPLETION_ITEM_SELECTED } from "./constants";
 import { IMavenCompletionItemProvider } from "./IArtifactProvider";
 import { getArtifacts, getVersions } from "./requestUtils";
 import { getSortText } from "./versionUtils";
-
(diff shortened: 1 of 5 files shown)


src/completion/centralProvider.ts
```diff
@@ -7,6 +7,8 @@ import { IMavenCompletionItemProvider } from "./IArtifactProvider";
 import { getArtifacts, getVersions } from "./requestUtils";
 import { getSortText } from "./versionUtils";
 
+const COMMAND_COMPLETION_ITEM_SELECTED: string = "maven.completion.selected";
```
Eskibear: the constant command name should be put somewhere else.

src/completion/centralProvider.ts
```diff
@@ -17,6 +19,10 @@ class CentralProvider implements IMavenCompletionItemProvider {
             const item: vscode.CompletionItem = new vscode.CompletionItem(gid, vscode.CompletionItemKind.Module);
             item.insertText = gid;
             item.detail = "central";
+            item.command = {
+                title: "selected", command: COMMAND_COMPLETION_ITEM_SELECTED,
```
Eskibear: the object can be created outside the loop to reduce memory usage

PR #240, opened 2019-01-31
Inject custom envs on creating terminal

Eskibear:
No need to compose the commands to set envs.


Eskibear:
@jdneo I'd like to also update the changelog in this PR.

jdneo:
Please go ahead

PR #247, opened 2019-02-13
Use workspaceFolder as the default target folder when creating projects

Eskibear:
The command `Generate from Maven Archetype` can be triggered from:
- a) Command palette.
- b) Context menu of a folder. 

For a), a default location is missing, this PR use the first workspace folder (if there is) as the default target folder.

### DIFF
--- a/src/archetype/ArchetypeModule.ts
+++ b/src/archetype/ArchetypeModule.ts
@@ -4,7 +4,7 @@
 import * as fse from "fs-extra";
 import * as os from "os";
 import * as path from "path";
-import { Uri, window } from "vscode";
+import { Uri, window, workspace } from "vscode";
 import { instrumentOperationStep, sendInfo } from "vscode-extension-telemetry-wrapper";
 import { OperationCanceledError } from "../Errors";
 import { getPathToExtensionRoot } from "../utils/contextUtils";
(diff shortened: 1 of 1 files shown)


src/archetype/ArchetypeModule.ts
```diff
@@ -55,7 +55,13 @@ export namespace ArchetypeModule {
         sendInfo(operationId, { archetypeArtifactId: artifactId, archetypeGroupId: groupId });
 
         // choose target folder.
-        const cwd: string = await instrumentOperationStep(operationId, "chooseTargetFolder", chooseTargetFolder)(entry);
+        let targetFolderHint: Uri;
+        if (entry) {
+            targetFolderHint = entry;
+        } else if (workspace.workspaceFolders.length > 0) {
+            targetFolderHint = workspace.workspaceFolders[0].uri;
```
jdneo: Do we need to take the multi-root into consideration?

Eskibear: How to? I can think of two ways: 1) ask users to select a root. 2) infer from active file.

1) introduces an extra step. 2) doesn't make much sense and you need to handle case where no active file is open. Meanwhile, it's possible that user want to create a project anywhere else. It's just a default location for the file selector. 

My point is, it's better than previous behavior without adding extra steps. So I prefer not to support that now.

Issue #253, opened 2019-02-19
shortcut command to add dependency in current pom.xml


Eskibear:
When the active file is pom.xml, add a shortcut to query Maven central repo and add dependencies. Steps are like:
- execute a command e.g. "Add new dependency"
- require to input keywords/artifactId/groupId/etc.
- query Maven central by the input, list available artifacts (gid:aid).
- select one artifact
- fetch all available versions, require to select.
- select one version
- it auto generate code blocks and insert into the pom.xml

PR #258, opened 2019-02-22
add shortcut command to add dependency in current pom.xml

Eskibear:
For #253

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -143,9 +143,5 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
     context.subscriptions.push(vscode.languages.registerCompletionItemProvider(pomSelector, completionProvider, ".", "-"));
     registerCommand(context, "maven.completion.selected", sendInfo, true);
 
-    registerCommand(context, "maven.project.addDependency", async () => {
-        if (vscode.window.activeTextEditor) {
-            await addDependencyHandler(vscode.window.activeTextEditor.document);
-        }
-    });
+    registerCommand(context, "maven.project.addDependency", async () => await addDependencyHandler());
 }
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -145,7 +145,7 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
 
     registerCommand(context, "maven.project.addDependency", async () => {
         if (vscode.window.activeTextEditor) {
-            await addDependencyHandler(vscode.window.activeTextEditor.document.uri.fsPath);
+            await addDependencyHandler(vscode.window.activeTextEditor.document);
         }
     });
 }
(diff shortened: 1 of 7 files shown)


src/handlers/addDependencyHandler.ts
```diff
@@ -0,0 +1,101 @@
...
+    return [
+        "",
+        "<dependencies>",
+        `${indent}<dependency>`,
+        `${indent}${indent}<groupId>${gid}</groupId>`,
+        `${indent}${indent}<artifactId>${aid}</artifactId>`,
+        `${indent}${indent}<version>${version}</version>`,
+        `${indent}</dependency>`,
+        "</dependencies>"
+    ].join(`\n${baseIndent || ""}${indent}`);
```
jdneo: `\n` or `os.EOL`?

src/extension.ts
```diff
@@ -137,7 +138,14 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
...
     );
+    const pomSelector: vscode.DocumentSelector = [{ language: "xml", scheme: "file", pattern: "**/pom.xml" }];
     // completion item provider
-    context.subscriptions.push(vscode.languages.registerCompletionItemProvider([{ language: "xml", scheme: "file", pattern: "**/pom.xml" }], completionProvider, ".", "-"));
+    context.subscriptions.push(vscode.languages.registerCompletionItemProvider(pomSelector, completionProvider, ".", "-"));
     registerCommand(context, "maven.completion.selected", sendInfo, true);
+
+    registerCommand(context, "maven.project.addDependency", async () => {
+        if (vscode.window.activeTextEditor) {
+            await addDependencyHandler(vscode.window.activeTextEditor.document);
```
jdneo: Do we really need to pass the document here? since it can be get from everywhere as long as we import `vscode`

src/handlers/addDependencyHandler.ts
```diff
@@ -0,0 +1,101 @@
...
+
+function constructDependencyNode(gid: string, aid: string, version: string, indent: string = ""): string {
+    return [
+        "",
+        "<dependency>",
+        `${indent}<groupId>${gid}</groupId>`,
+        `${indent}<artifactId>${aid}</artifactId>`,
+        `${indent}<version>${version}</version>`,
+        "</dependency>"
+    ].join(`\n${indent || ""}${indent}`);
```
jdneo: `\n` or `os.EOL`?

jdneo: Oh, not `os.EOL`, `window.activeTextEditor!.document.eol`

src/extension.ts
Eskibear: Indeed, no need to set a param. And the "add a dependency" command only serves for the current active document, it's another point that we don't need a param. 

The reason I pass the document as the param is, in future there's possibility to "add a dependency" for a project without opening its pom.xml. But for the moment, it doesn't make much sense.

src/handlers/addDependencyHandler.ts
Eskibear: now use `eol === LF ? "\n" : "\r\n"`

src/extension.ts
Eskibear: param document removed.

Issue #260, opened 2019-02-26
Hover on dependency to show effective version


Eskibear:
Often, the version is managed by "dependencyManagement". And version conflict is also common. So it's convenient to view the effective version of a dependency. 

The basic idea is, if we have effective pom calculated, it's easy to provide hover information from that.

PR #262, opened 2019-02-26
Hover to see effective version of dependency

Eskibear:
Only works after effective pom has been calculated. 

For #260


Eskibear:
FYI, it looks as below: 

image

src/hover/hoverProvider.ts
```diff
@@ -0,0 +1,57 @@
...
+                            `version = ${effectiveVersion}`
+                        ].join("\n\n"), targetRange);
+                    }
+                }
+            }
+            default:
+                return undefined;
+        }
+    }
+
```
jdneo: nit: remove the empty line here

PR #281, opened 2019-03-25
Enable debugging plugin goal

Eskibear:
1. create a terminal with debugging jvm args injected by MAVEN_OPTS
2. execute mvn goal
3. attach debugger

### DIFF
--- a/package.nls.json
+++ b/package.nls.json
@@ -26,5 +26,4 @@
     "configuration.maven.terminal.favorites": "Specify pre-defined favorite commands to execute.",
     "configuration.maven.terminal.favorites.alias": "A short name for the command.",
     "configuration.maven.terminal.favorites.command": "Content of the favorite command."
-    
 }
--- a/package.nls.json
+++ b/package.nls.json
@@ -9,6 +9,7 @@
     "contributes.commands.maven.history": "History ...",
     "contributes.commands.maven.goal.execute": "Execute commands",
     "contributes.commands.maven.plugin.execute": "Execute",
+    "contributes.commands.maven.plugin.debug": "Debug",
     "contributes.commands.maven.view.hierarchical": "Switch to hierarchical view",
     "contributes.commands.maven.view.flat": "Switch to flat view",
     "contributes.commands.maven.project.addDependency": "Add a dependency",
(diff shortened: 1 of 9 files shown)


package.nls.json
```diff
@@ -25,4 +26,5 @@
     "configuration.maven.terminal.favorites": "Specify pre-defined favorite commands to execute.",
     "configuration.maven.terminal.favorites.alias": "A short name for the command.",
     "configuration.maven.terminal.favorites.command": "Content of the favorite command."
+
```
jdneo: Redundant empty line

src/handlers/debugHandler.ts
```diff
@@ -0,0 +1,41 @@
...
+async function debug(pomPath: string, command: string): Promise<void> {
+    const freePort: number = await getPort();
+    const mavenOpts: string = [
+        Settings.getEnvironment().MAVEN_OPTS, // user-setting MAVEN_OPTS
+        `-Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=${freePort}` // MAVEN_DEBUG_OPTS
+    ].filter(Boolean).join(" ");
+    const sessionId: string = createUuid().substr(0, 6);
+    const debugTerimnal: vscode.Terminal = await executeInTerminal(command, pomPath, {
+        name: `mvnDebug ${sessionId}: ${command}`,
+        env: { MAVEN_OPTS: mavenOpts }
```
jdneo: Push the `process.env` into the `env` here as well?

src/handlers/debugHandler.ts
```diff
@@ -0,0 +1,41 @@
...
+    });
+    const debugConfig: vscode.DebugConfiguration = {
+        type: "java",
+        name: `Maven (Attach) - ${sessionId}`,
+        request: "attach",
+        hostName: "localhost",
+        port: freePort,
+        terminalName: debugTerimnal.name
```
jdneo: Need to provide project name

Eskibear: Just discussed with debugger extension maintainer, I would rather leave it empty, debugger will try resolving it.

src/handlers/debugHandler.ts
Eskibear: on creating a terminal, vscode already ensures that. 
```ts
		/**
		 * Object with environment variables that will be added to the VS Code process.
		 */
		env?: { [key: string]: string | null };
```

PR #288, opened 2019-03-29
Add TS strict null check

Eskibear:
TL;DR


tsconfig.json
```diff
@@ -4,6 +4,7 @@
         "noUnusedParameters": true,
         "noImplicitAny": true,
         "noImplicitReturns": true,
+        "strictNullChecks": true,
```
Eskibear: All other changes are because of this line.

PR #298, opened 2019-04-05
Fix a typo in TestPlan

balloonio:
The setting config example is wrong. I came across this example when I was trying to set up my maven favorite commands. I got errors until I realize the correct setting field should be "maven.terminal.favorite**s**"

### DIFF
--- a/TestPlan.md
+++ b/TestPlan.md
@@ -57,7 +57,7 @@
 1. Specify a favorite command in settings, e.g.
     ```
         {
-            "maven.terminal.favorite": [
+            "maven.terminal.favorites": [
                 {
                     "alias": "full-build without tests",
                     "command": "clean package -DskipTests"


tsiq-bolunz:
Also, it might be a good idea to add to the README.md how to create a customized favorite command, as the current README.md has no mentioning of how to do that at all.

balloonio:
Also, it might be a good idea to add to the README file how to add a customized maven command to favorites, not just in the TestPlan file. This feature is very useful and the current README file has no mentioning of that at all.

balloonio:
@Eskibear

Eskibear:
Thank you for the contribution. Are you willing also to update the README? Suggest to add into this section https://github.com/Microsoft/vscode-maven/blob/master/README.md#additional-configurations

PR #366, opened 2019-07-15
Migrate to use vscode-test

Eskibear:


src/test/runTest.ts
```diff
@@ -0,0 +1,23 @@
+import * as path from "path";
```
jdneo: license

src/test/suite/index.ts
```diff
@@ -0,0 +1,38 @@
+import * as glob from "glob";
```
jdneo: license

PR #395, opened 2019-09-17
Use constant field to represent code value

RunxinXu:
The constant field is from: https://help.eclipse.org/2019-06/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Fconstant-values.html

### DIFF
--- a/src/artifactSearcher.ts
+++ b/src/artifactSearcher.ts
@@ -8,7 +8,12 @@ import { registerCommand } from "./extension";
 import { executeJavaLanguageServerCommand } from "./jdtls/commands";
 import { applyWorkspaceEdit } from "./utils/editUtils";
 
-const unresolvedCode: string[] = ["16777218", "570425394"];
+// Please refer to https://help.eclipse.org/2019-06/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Fconstant-values.html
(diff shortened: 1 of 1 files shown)


src/artifactSearcher.ts
```diff
@@ -8,7 +8,12 @@ import { registerCommand } from "./extension";
 import { executeJavaLanguageServerCommand } from "./jdtls/commands";
 import { applyWorkspaceEdit } from "./utils/editUtils";
 
-const unresolvedCode: string[] = ["16777218", "570425394"];
+// Please refer to https://help.eclipse.org/2019-06/index.jsp?topic=%2Forg.eclipse.jdt.doc.isv%2Freference%2Fapi%2Fconstant-values.html
+// tslint:disable-next-line: variable-name
```
Eskibear: follow tslint rule to rename it as `UNDEFINED_TYPE` etc

Eskibear: or create a enum `CategorizedProblem`, e.g. 
```
enum CategorizedProblem {
UNDEFINED_TYPE = "xxxx",
...
}
```

RunxinXu: I thought using the same name as that of eclipse document would be better. seems not to be what I thought ahhh.

Issue #410, opened 2019-10-15
Feature request


scandinave:
## Suggestion

Add an option to use archetype from a custom private repo like nexus.
## Use Cases
Working offline, i'm unable to generate project from archetype inside VSCode

polardegausser:
This one would be very useful. I just built a custom archetype and tried using it from scratch (removing it from .m2 and pulling it from nexus), but then I realized I can't add custom archetype catalog URL.

Eskibear:
A setting (e.g. `maven.archetype.catalogUrl`) would work, with a default value of the maven central. 

Note that according to current implementation, metadata of archetypes are retrieved and saved to a local .json file. After you change the value, you have to run "Update archetype catalog" to fetch and generate the metadata. And that means you cannot list archetypes of  two catalogs together (e.g. maven central and your private nexus repo).

I imagine that a setting maven.archetype.catalogUrl*s* would be more help, if possible.

Eskibear:
First thank @scandinave for creating PR #412 , providing the solution with the new setting mentioned above. 

After I add a nexus repo into the vscode settings, it does list the archetypes from a custom catalog. But I cannot simply create a project because the nexus repo is not specified in my settings.xml. And if there's inconsistency between vscode settings and Maven settings.xml, things would get wired, especially for a user who's not as experienced as you guys.

So I'm thinking if we should read the available catalogs from Maven settings.xml, e.g. by executing `mvn help:effective-settings`. So there would be only one source of truth. 
@polardegausser @scandinave what do you guys think? which one would be more helpful?

polardegausser:
Hi,

I created the PR :) I think that the approach with Maven's settings.xml file is a good path to follow. I did the #412 as a quickfix since I needed this feature for my job and I definitely don't want to fallback to Eclipse just because of this missing link. I like the simple workflow I set up for myself in VSCode for Java.

I can try implementing this by iterating through repositories from settings.xml (I assume most of the logic for doing this is already available in the current source) and downloading archetype catalog from every [repo_url]/archetype-catalog.xml.
This way users won't need to specify anything separately, instead we can rely on maven. Of course, the question is then, does the user want to see archetypes from all of the remote repositories? For example, I might have some additional repos added for dependencies but not for their archetypes, the list might get very large.

scandinave:
@Eskibear i think only one source of truth is the right way. But this need to be well documented, so user can quickly understand how this work.

Eskibear:
[...]

PR #411, opened 2019-10-18
Custom archetype catalogs

polardegausser:
#410 Added support for custom archetype catalogs through settings.json file

Please review this code and decide if it is a good solution for the referenced issue.


polardegausser:
Fixed code to pass tslint checks. Closing this pull request and creating a new one from current commit.

PR #412, opened 2019-10-18
Custom archetype catalogs

criskurtin:
#410 Added support for custom archetype catalogs through settings.json file

Please review this code and decide if it is a good solution for the referenced issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,6 +178,26 @@ Now right-click on an project item, and then click `Favorite ...`. The option `f
 
 </details>
 
+<details>
+<summary>Customize remote Maven archetype catalogs</summary>
+
+Specify a remote archetype catalog in settings:
+
+```json
+{
+    "maven.archetype.catalogs": [
+        {
+            "name": "Contoso catalog",
+            "url": "http://nexus.contoso.com/service/local/repositories/releases/content/archetype-catalog.xml"
+        }
+    ]
+}
+```
+
+Now open up the Command Palette -> select `Maven: Update Maven Archetype Catalog`.
+
+</details>
+
 ## Settings
 
 | Name | Description | Default Value |
--- a/package.nls.zh.json
+++ b/package.nls.zh.json
@@ -28,5 +28,8 @@
     "configuration.maven.view": "指定 Maven 项目的视图方式。",
     "configuration.maven.terminal.favorites": "指定要执行的偏好命令。",
     "configuration.maven.terminal.favorites.alias": "命令的简称。",
-    "configuration.maven.terminal.favorites.command": "命令的内容。"
+    "configuration.maven.terminal.favorites.command": "命令的内容。",
+    "configuration.maven.archetype.catalogs": "Specify custom remote catalogs from which to fetch archetypes.",
+    "configuration.maven.archetype.catalogs.name": "A name of the remote catalog.",
+    "configuration.maven.archetype.catalogs.url": "URL from which to download the remote catalog."
 }
--- a/src/archetype/ArchetypeModule.ts
+++ b/src/archetype/ArchetypeModule.ts
@@ -94,12 +94,14 @@ export namespace ArchetypeModule {
         const targetFilePath: string = path.join(getPathToExtensionRoot(), "resources", "archetypes.json");
         const catalogs = Settings.Archetype.catalogs();
         let archetypes: Archetype[] = [];
-        let tempArchetypes: Archetype[] = await parseArchetypeCatalog(REMOTE_ARCHETYPE_CATALOG_URL);
-        archetypes.concat(tempArchetypes);
+        await parseArchetypeCatalog(REMOTE_ARCHETYPE_CATALOG_URL).then(response => {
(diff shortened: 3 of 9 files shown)


package.nls.zh.json
```diff
@@ -28,5 +28,8 @@
     "configuration.maven.view": "指定 Maven 项目的视图方式。",
     "configuration.maven.terminal.favorites": "指定要执行的偏好命令。",
     "configuration.maven.terminal.favorites.alias": "命令的简称。",
-    "configuration.maven.terminal.favorites.command": "命令的内容。"
+    "configuration.maven.terminal.favorites.command": "命令的内容。",
+    "configuration.maven.archetype.catalogs": "Specify custom remote catalogs from which to fetch archetypes.",
+    "configuration.maven.archetype.catalogs.name": "A name of the remote catalog.",
+    "configuration.maven.archetype.catalogs.url": "URL from which to download the remote catalog."
```
polardegausser: Requires a correct translation

polardegausser:
Default hardcoded string is left in the code. This might be pulled out to the settings so it is modifiable but for now it is probably better to leave it in there than not having anything listed when creating a new project.

Eskibear:
Thank you for contributing, and I just played for a little while, it did list the archetypes from a custom catalog.
But I cannot simply successfully create a project because the assembled command didn't specify the catalog.

```log
> cmd /c mvn archetype:generate -DarchetypeArtifactId="yanzh" -DarchetypeGroupId="com.microsoft.azure" -DarchetypeVersion="1.10"
[INFO] Scanning for projects...
[INFO]
[INFO] ----------------------< com.microsoft:quickstart >----------------------
[INFO] Building quickstart 0.1
[INFO] --------------------------------[ jar ]---------------------------------
[INFO] 
[INFO] >>> maven-archetype-plugin:3.0.1:generate (default-cli) > generate-sources @ quickstart >>>
[INFO]
[INFO] <<< maven-archetype-plugin:3.0.1:generate (default-cli) < generate-sources @ quickstart <<<
[INFO]
[INFO]
[INFO] --- maven-archetype-plugin:3.0.1:generate (default-cli) @ quickstart ---
[INFO] Generating project in Interactive mode
[WARNING] Archetype not found in any catalog. Falling back to central repository.
...
```

 For an experienced Maven user, what's the convention/common practice to make use of a custom archetype catalog? Any idea to provide simple and straightforward workflow for users who are not that experienced?

Eskibear:
As far as I know, to get it work, you have to also add the repository in your maven settings.xml . And thus, the user should change both settings.xml and vscode settings. This is not a good practice, as we should keep only one source of truth if possible.

My proposal is that, the maven extension automatically reads effective maven settings, and fetch the archetypes from the user-defined repositories.

README.md
```diff
@@ -178,6 +178,26 @@ Now right-click on an project item, and then click `Favorite ...`. The option `f
...
+}
+```
+
+Now open up the Command Palette -> select `Maven: Update Maven Archetype Catalog`.
+
+</details>
+
 ## Settings
 
 | Name | Description | Default Value |
```
Eskibear: Add a row about the new setting here.

package.nls.zh.json
Eskibear: LGTM, you can just leave it here for others to contribute.

src/archetype/ArchetypeModule.ts
```diff
@@ -89,14 +93,25 @@ export namespace ArchetypeModule {
...
+            for (const catalog of catalogs) {
+                archetypes = archetypes.concat(await parseArchetypeCatalog(catalog.url));
+            }
+        }
         await fse.ensureFile(targetFilePath);
         await fse.writeJSON(targetFilePath, archetypes);
     }
 
+    async function parseArchetypeCatalog(url: string): Promise<Archetype[]> {
+        const xml: string = await Utils.downloadFile(url, true);
```
Eskibear: If the downloading fails, e.g. due to a wrong url or networking issue, an exception will be thrown. Thus even though the latest list of Maven Central archetypes is successfully retrieved, it's not updated into the local JSON file. Is that the behavior you want? 
* You thrown the expection to inform user about the error?
or 
* You swallow the exception and return an empty list for the malformed catalog silently?

criskurtin:
Should we close this PR? This branch went stale a long time ago. A fresh attempt should be made.

Eskibear:
Agree. Let's close it and create a new one.

BTW instead of introducing a new setting, the ideal solution in my mind would be, to read repository information from settings.xml, and automatically fetch the metadata when updating.

E.g. settings.xml
```
  <repository>
    <id>archetype</id>
    <url>https://repository.domain.com/path/to/repo/</url>
  </repository>
```

And another concern is, when we choose archetypes from custom repository, we may need to specify `archetypeCatalog` in the command line. See https://maven.apache.org/archetype/maven-archetype-plugin/generate-mojo.html

criskurtin:
good points. closing this PR, but it should definitely be referenced in a new one.

PR #435, opened 2019-12-03
refactor: extract logic into different handlers

Eskibear:


src/mavenTerminal.ts
```diff
@@ -73,11 +58,15 @@ class MavenTerminal implements vscode.Disposable {
         }
     }
 
-    public dispose(id?: string): void {
-        if (id) {
-            this.terminals[id].dispose();
+    public dispose(terminalName?: string): void {
+        if (terminalName) {
```
jdneo: if (this.terminals[terminalName])

Eskibear: done

PR #457, opened 2020-01-20
allow to browse for local maven binary if not found

Eskibear:
browse-mvn

### DIFF
--- a/src/utils/mavenUtils.ts
+++ b/src/utils/mavenUtils.ts
@@ -93,7 +93,7 @@ export async function executeInTerminal(options: {
     env?: { [key: string]: string };
     terminalName?: string;
 }): Promise<vscode.Terminal | undefined> {
-    const {command, mvnPath, pomfile, cwd, env, terminalName} = options;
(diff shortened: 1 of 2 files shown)


src/utils/mavenUtils.ts
```diff
@@ -170,9 +170,50 @@ function getTempTolder(identifier: string): string {
...
+    const selectedUris: vscode.Uri[] | undefined = await vscode.window.showOpenDialog({
+        openLabel: `Select ${mvnFilename}`,
+        canSelectMany: false,
+        filters
+    });
+    if (selectedUris === undefined) {
+        return undefined;
     }
+
+    const mvnPath: string | undefined = selectedUris.length > 0 && selectedUris[0] !== undefined ? selectedUris[0].fsPath : undefined;
```
jdneo: const mvnPath: string | undefined = selectedUris[0]?.fsPath;

Eskibear: tried but failed the tslint. let's leave the concise way in future PRs.

Issue #474, opened 2020-03-05
Resolve unknown fype will show "Command 'maven.artifactSearch'" fails.


Elvira-sxr:
**Describe the bug**
Resolve unknown fype will show "Command 'maven.artifactSearch'" fails.

**To Reproduce**
1. Open a Maven project in VS Code, and open a .java file. In body of a function, type below statements.
```
Gson gson;
StringUtils.capitalize("abc");
```
2.  Open Code Actions, "Resolve unknown type" is in the list.
3. Clicking "Resolve unknown type", it will show error message.

![image](https://user-images.githubusercontent.com/51812626/75973766-b0e2f200-5f10-11ea-94f0-e8012afb8950.png)

**Environments:**
 - OS: Win10
 - VS Code version: 1.42.1
 - VSCode language server version : 0.58.0
 - Maven for java version : 0.21.0

Eskibear:
I can reproduce it **iff.** I'm using Java 13

Eskibear:
```
[Error - 11:46:43 AM] Mar 6, 2020, 11:46:43 AM Error in calling delegate command handler
org/codehaus/plexus/configuration/PlexusConfiguration
java.lang.NoClassDefFoundError: org/codehaus/plexus/configuration/PlexusConfiguration
	at org.apache.maven.index.ArtifactInfo.<clinit>(ArtifactInfo.java:138)
	at org.apache.maven.index.FlatSearchRequest.<init>(FlatSearchRequest.java:50)
	at com.microsoft.java.maven.BaseClassSearcher.search(ArtifactSearcher.java:381)
	at com.microsoft.java.maven.BaseClassSearcher.searchByClassName(ArtifactSearcher.java:339)
	at com.microsoft.java.maven.ClassSearcher.searchByClassName(ArtifactSearcher.java:213)
	at com.microsoft.java.maven.ArtifactSearcher.searchByClassName(ArtifactSearcher.java:80)
	at com.microsoft.java.maven.handler.DelegateCommandHandler.executeCommand(DelegateCommandHandler.java:35)
	at org.eclipse.jdt.ls.core.internal.handlers.WorkspaceExecuteCommandHandler$1.run(WorkspaceExecuteCommandHandler.java:215)
	at org.eclipse.core.runtime.SafeRunner.run(SafeRunner.java:45)
[...]

PR #475, opened 2020-03-11
Fix ClassNotFound error

Eskibear:
`PlexusConfiguration` is included in `org.eclipse.m2e.maven.runtime`


Eskibear:
this possibly fixes #474

Eskibear:
verified with pr build.

PR #488, opened 2020-04-16
Hide create maven archetype command in palette if dependency viewer enabled

jdneo:
Signed-off-by: Sheng Chen <sheche@microsoft.com>

### DIFF
--- a/package.json
+++ b/package.json
@@ -265,6 +265,10 @@
         {
           "command": "maven.goal.execute.fromProjectManager",
           "when": "never"
+        },
+        {
+          "command": "maven.archetype.generate",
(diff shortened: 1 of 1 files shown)


jdneo:
@Eskibear Do you think the context value should be set at the dependency viewer side?

Eskibear:
This makes me think of a soft extension dependency, using `javaDependencyViewer:activated`, it's easy to check whether an extension is activated or not. But that's not robust (comparing to the extensionAPI mechanism) if you set the context value early before the `activate` function completes.

Eskibear (approved):

Issue #489, opened 2020-04-28
Maven output does not  write to the VSC terminal window


RedWingBB:
When I run maven/package a new cmd window is launched and the build output is written to it

I right click the maven project, then Package
A cmd window pops up
The build output streams into it
Then the cmd window closes
This makes it difficult to see if the build succeeded, what were the problems, so on

It may be there is a configuration option in the plugin or in VSC itself that I havent found

Environment is an admin login on  Windows Server 2090
Uploading a screen shot, I hope it shows up.

![image](https://user-images.githubusercontent.com/41631548/80506344-71021f00-8943-11ea-868f-4fcf3fde4bae.png)

Eskibear:
The command is expected to run on integrated terminal withing vscode. 
-  What's your vscode's default integrated terminal?

Eskibear:
@RedWingBB 
It's probably because you've wronly set `maven.executable.path` by your own.  Please check if the path ends with "mvn.cmd" instead of "mvn".

RedWingBB:
Thanks for the reply
Here is my command line
C:\Program Files (x86)\Apache\apache-maven-3.6.3\bin\mvn
My Terminal is Powershell

Eskibear:
If the ".cmd" file extension is not specified, powershell will run a CMD propmt in a new window. 

#### What you can do now
Append ".cmd" to "mvn", that it, to set `maven.executable.path` to `C:\Program Files (x86)\Apache\apache-maven-3.6.3\bin\mvn.cmd`. Have a try, and let me know if it works.

Eskibear:
Fixed in #499

RedWingBB:
Adding .cmd did it
Thanks for the reply

Nicryc:
@Eskibear Since the last update of VSCode (August 2021 – 1.60) I'm facing this issue too. 

Maven is forced inside a Powershell __integrated__ terminal (although my default terminal is cmd) and run the following command : `& "path\to\apache-maven-3.2.5\bin\mvn" clean install -f "path\to\pom.xml"` which has for effect to open an __external__ cmd terminal and running Maven inside. Everything was working as intended before this update though, that's really weird.

Adding a `.cmd` to `mvn` just give me a `CommandNotFoundException`

Working on Windows 10. Executed mvn through a right click to the Maven project and used a favourite command defined in `maven.terminal.favorites` : `clean install`

Eskibear:
[...]

PR #494, opened 2020-05-20
Open cmd when running inside PowerShell

arpontes:
As of today, when VSCode default's terminal is PowerShell, running a maven command opens a new cmd window, and when the process is finished, this window is closed and we can't see what happened.

With this change, we force PowerShell to run the command in it's own window.

### DIFF
--- a/src/mavenTerminal.ts
+++ b/src/mavenTerminal.ts
@@ -79,7 +79,7 @@ function getCommand(cmd: string): string {
     if (process.platform === "win32") {
         switch (currentWindowsShell()) {
             case WindowsShellType.POWERSHELL:
-                return `& ${cmd}`; // PowerShell
+                return `cmd /c ${cmd}`; // PowerShell
             default:
                 return cmd; // others, try using common one.
         }


Eskibear:
Also reported in #489 , but I was not able to reproduce the bug. 
In my observation, with "cmd /c", powershell just executed the command in the integrated terminal, without poping up new a window.  Can you share the version of your powershell? 

BTW, the "cmd /c" trick was to cover some corner case (if I don't remember wrong). I'll let you know when I find out the back-story.

arpontes:
I think I found something that could help to understand the issue.

In my settings.json, the configurations relative to java and maven are:
    "java.home": "D:\\java\\jdk-14.0.1",
    "maven.executable.path": "D:\\java\\plugins\\maven\\lib\\maven3\\bin\\mvn",
    "maven.terminal.useJavaHome": true

This way, a new cmd window pops out and closes as soon as the execution ends, therefore I can't see what happened.

I found out that when I reference the .cmd version of mvn, the results are printed to PS terminal window, without opening a new one.
    "maven.executable.path": "D:\\java\\plugins\\maven\\lib\\maven3\\bin\\mvn.cmd"

So, that's why the "cmd /c" works! When running "cmd /c", even if we don't put .cmd, it reached for mvn.cmd.

Eskibear:
I find the story. Before PR #455 , we do have "cmd /c" . You may take a look at it if you have interest.  Anyway, now since it's breaking your case, I agree that we add it back.

Eskibear:
@arpontes
The "cmd /c" solution doesn't work for some corner cases. Running a regression test, we find that if you have whitespaces in the path, the command will not be correctly parsed by powershell. I think that's the reason why we swtiched to "&" solution long time ago.

Screen Shot 2020-05-26 at 3 50 56 PM


To fix the issue you mentioned, what I'm thinking now is:
- revert this PR to use "&" solution.
- on windows, once we detect `maven.executable.path` is set to "path/to/mvn", we append ".cmd" before executing it.

arpontes:
@Eskibear actually, the problem is the enclosing quotes.
This seems like a cmd bug to me, but the problem is: when we call cmd /c, the entire sequence should be evaluated by cmd, and the parameters should be passed to the executable (or batch) in the first parameter.

I created a series of folders with an .cmd file that contains only a "dir".
d:\dev\teste.cmd
d:\dev\plugins test\teste.cmd
(and much more)
All of then runned fine.

Then I start to test some parameters, and the problem arose:
cmd /c "D:\test\plugins test\teste.cmd" "x y"
See, the problem cames when the batch file is in a folder with space AND there's another parameter with a space.

The problem is solved if we enclose all the command in quotes (scaping the original quotes):
cmd /c """D:\test\plugins test\teste.cmd"" ""d:\deva test\"""

The same worked with maven:
cmd /c """D:\java\plugins test\maven\lib\maven3\bin\mvn.cmd"" clean -f ""D:\dev test\almundo\api-self-management-reservations\pom.xml"""

Actually, this was the right way to call cmd in order to be a single command - including the parameters
... :-)
But if you already have done the change you suggested, I think it's fine too.

Eskibear:
Thanks for the information, and it helps me understand more about the behavior. 
For the moment, if the issue can be addressed by forcing to use mvn.cmd on Windows, I think it's more wise to make minimal changes, instead of using double/triple quotes. 
We'll revisit your proposal and refer to official docs/guides of powershell if new problem comes.

PR #499, opened 2020-05-26
always append .cmd for mvn executable on Windows

Eskibear:
See https://github.com/microsoft/vscode-maven/pull/494#issuecomment-633869294


Eskibear:
Actually I'm concerning that if whether it affects "Git Bash" or "WSL" on windows...

Eskibear:
Now I only append .cmd for powershell, to make minimal impact.

Eskibear:
@jdneo please take a look ;-)

PR #511, opened 2020-07-06
update imports and types of module "xml-zero-lexer"

943297456:


.gitignore
```diff
@@ -15,3 +15,9 @@ resources/IndexData/index/*.cfe
 resources/IndexData/index/*.cfs
 resources/IndexData/index/*.si
 resources/IndexData/index/segments*
+
+.vs/*
```
Eskibear, line 19: what for?

Eskibear:
Suggest that you directly help to update and publish `@types/xml-zero-lexer`. So we can import it in TS code without the duplicate definition.

.gitignore
943297456: some files are generated by Visual Studio IDE

Eskibear:
I'm ok to merge this PR with below asks:
* can you remove `.vscode` from `.gitignore`, as we do want to include vscode-related files, including whatever will be in that folder in the future. 
* do you have the interest to update and publish `@types/xml-zero-lexe`?

.gitignore
Eskibear: ok jusjt keep it, but please remove the `.vscode` part.

943297456:
> Suggest that you directly help to update and publish `@types/xml-zero-lexer`. So we can import it in TS code without the duplicate definition.


That's a good idea.

I'll take a moment to learn more about this module and provide its definition.

PR #513, opened 2020-07-10
Fix error: When a Non-Maven terminal is closed, all Maven terminals are shut down.

943297456:
Fix error: When a Non-Maven terminal is closed, all Maven terminals are shut down.

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -83,10 +83,13 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
 
     registerConfigChangeListener(context);
 
-    // Free resources when manually closing a terminal
+    // Free resources when a terminal is manually closed
     context.subscriptions.push(
         vscode.window.onDidCloseTerminal((closedTerminal: vscode.Terminal) => {
-            mavenTerminal.dispose(closedTerminal.name);
+            const name = mavenTerminal.find(closedTerminal);
+            if (name !== undefined) {
+                mavenTerminal.dispose(name);
+            }
         })
     );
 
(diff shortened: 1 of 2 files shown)


Eskibear:
> Fix error: When a Non-Maven terminal is closed, all Maven terminals are shut down.

I cannot reproduce it. Can you provide more detailed steps?

943297456:
> > Fix error: When a Non-Maven terminal is closed, all Maven terminals are shut down.
> 
> I cannot reproduce it. Can you provide more detailed steps?

First:
	Create one or more terminals by the Command[Maven:Execute Commands] and I call them Maven terminals;
	Create one or more terminals by the Command[Terminal:Create New Integrated Terminal] and I call them Non-Maven terminals.
Then:
	Switch to a Non-Maven terminal and close it by the Command[Terminal:Kill the Active Terminal Instance],
	You can see that all Maven terminals are closed.

Eskibear:
Good catch. Instead of putting the logic here, you should modify the listener in https://github.com/microsoft/vscode-maven/blob/4fbe876b053ecba5503420f1681852caff427619/src/extension.ts#L89

Check the `closedTerminal`, if it's a non-maven terminal, do not call `mavenTerminal.dispose()`

Eskibear:
> ERROR: (typedef) /home/travis/build/microsoft/vscode-maven/src/extension.ts[89, 19]: expected variable-declaration: 'name' to have a typedef

`npm run tslint` before you push your commit.

src/extension.ts
```diff
@@ -109,9 +112,9 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
 
     // debug
     registerCommand(context, "maven.plugin.debug", debugHandler);
-    vscode.debug.onDidTerminateDebugSession((session: any) => {
+    vscode.debug.onDidTerminateDebugSession((session) => {
         if (session.type === "java") {
-            const terminalName: string = session._configuration.terminalName;
+            const terminalName: string = session.configuration.terminalName;
```
Eskibear, line 117: data structure of `session` changes?

943297456: > data structure of `session` changes?

'configuration' is the get method of '_configuration' and it is provided in the type definition, so I used it instead.

PR #531, opened 2020-08-26
register 'execute commands' onto Project Manager

Eskibear:
Requires https://github.com/microsoft/vscode-java-dependency/pull/311

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -746,8 +746,7 @@
                 },
                 "kind-of": {
                     "version": "6.0.2",
-                    "resolved": "https://registry.npmjs.org/kind-of/-/kind-of-6.0.2.tgz",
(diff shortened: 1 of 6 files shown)


Eskibear:
Note that the wording "execute commands..." here might be not clear enough. 
In Maven extension, the full name of the command (in command palette) is "Maven: execute commands...", because the `group` is "Maven".

jdneo:
Will you consider change it to `Execute Maven Commands...`? IMO, trigger the command from Maven & Project explorer may be more frequency than in command palette.

Eskibear:
Then the side effect would be "Maven: Execute Maven Commands..." in command palette. And yes it's OK for me.

jdneo:
So now we have three possible solution per offline discussion:

- Change to `Maven: Execute Maven Commands...`
- Wrap another command for tree view usage
- Open a ticket to VS Code about adding a treeview display name for a command

Eskibear:
Now using an alias command `maven.goal.execute.fromProjectManager` for the same functionality. 

image

jdneo (approved):

Issue #534, opened 2020-09-04
Broken under WSL


soerenBoisen:
**Describe the bug**
I have my terminal in VS Code set to WSL 1 bash (Ubuntu).

1. When executing a Maven command without maven.executable.path in settings.json:
Extension attempts to execute mvnw.cmd instead of mvnw, leading to syntax errors.

2. When executing Maven command with `maven.executable.path: "/usr/share/maven/bin/mvn"`:
Extension wrongly adds Windows mount point to the absolute path:
```bash
-bash: /mnt/c/usr/share/maven/bin/mvn: No such file or directory
```

**Expected behavior**
With no Maven executable path set, I expect extension to execute mvnw in project, not mvnw.cmd, since the shell is bash and don't understand Windows batch files.

With Maven executable path set to an absolute path, I expect extension to run it without adding random and wrong mount point in front, since the path is supposed to be _absolute_.

**Environments (please complete the following information as much as possible):**
 - OS: Windows 10
 - VS Code version: 1.48.2
 - Extension version: 0.24.2

Eskibear:
Thanks for reporting this. I guess you are connecting to WSL via VS Code Remote extension. Just FYI that WSL support in this extension was added long before VS Code Remote is released, when WSL was only a different shell. At that time, `maven.executable.path` was **shared** between WSL and Windows, and the path was converted by executing `wslpath -u <path>`. That's why you see "/mnt/c/" prepended.

soerenBoisen:
I am unsure if I connect via VS Code Remote. My settings are also from long before Remote released 😄 

This is what I have in settings for terminal:
```json
    "terminal.integrated.shell.windows": "C:\\WINDOWS\\System32\\wsl.exe",
```

Eskibear:
Ok, then you are not using Remote. 
Here are two ways of using WSL:
* Only one maven distribution is installed in Windows (let's say `C:\Maven`), and the user wants to use this one in both Windows and WSL, so we have to convert `C:\Maven` to `/mnt/c/Maven/`. 
* Maven is installed in WSL and it's dedicated for WSL (e.g. `/usr/share/maven/` in your case). 

Currently I cannot think of a simple fix covering both cases above... You may want to try VS Code Remote if you prefer WSL, as that's also better supported by vscode itself.

soerenBoisen:
Why not simply check if the path starts with forward slash and don't perform conversion in that case?
[...]

PR #535, opened 2020-09-11
WSL: do not convert path if already unix-like

Eskibear:
improve robostness, see #534


src/mavenTerminal.ts
```diff
@@ -160,6 +160,10 @@ function toDefaultWslPath(p: string): string {
 }
 
 export async function toWslPath(filepath: string): Promise<string> {
+    if (filepath.startsWith("/")) { // unix-like absolute path
```
jdneo, line 163: How about `path.posix.isAbsolute()`

Eskibear: I don't see much difference for `path.posix.isAbsolute()`, it's also checking the first char. See https://github.com/nodejs/node/blob/d0377a825bf7ceb838570f434fdd7d4b1773b8fa/lib/path.js#L358-L359

I'm ok with both, any other benefit bisides better readability?

jdneo: Yes it's for the readability, the comment is no-need then.

jdneo (approved):

PR #536, opened 2020-09-13
Make optional the File Explorer context menu entry

tapeinosyne:
The globally enabled entry in the Explorer right-click menu is a teensy bit intrusive for users who are not working with Java/Maven exclusively or nearly so. This quick patch introduces an option to enable or disable the entry, and disables it by default.

A few notes:
  - Disabling the entry by default is bound to surprise a few users at the next update. Even so, I reckon that it should be the preferred behavior.
  - The informal "Generate project from maven archetypes" test from `TestPlan.md` is somewhat overloaded now that it also includes a procedure to test the menu setting. I'd be happy to split the test in two, one for the setting and one for project generation proper.
  - Line endings seemed to be inconsistent across files and, within `TestPlan.md`, across lines. I have tried not to alter them.
  - I am unable to provide a Chinese description for the setting. Feel free to provide me with one, open a PR against my branch, or make edits directly.

Also note that there is a similar issue with the "Maven" section for the Explorer, which is introduced by this extension and active by default. It is not addressed here because, while intrusive, it can be disabled with ease—unlike the context menu entry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -270,6 +270,7 @@
       "explorer/context": [
         {
           "command": "maven.archetype.generate",
+          "when": "config.maven.showInExplorerContextMenu && explorerResourceIsFolder",
           "group": "maven@3"
         },
         {
@@ -522,6 +523,12 @@
             "default": "flat",
             "scope": "window",
             "description": "%configuration.maven.view%"
+          },
+          "maven.showInExplorerContextMenu": {
+            "type": "boolean",
+            "default": "true",
+            "description": "%configuration.maven.showInExplorerContextMenu%",
+            "scope": "resource"
           },
           "maven.pomfile.globPattern": {
             "type": "string",
--- a/package.nls.json
+++ b/package.nls.json
@@ -27,6 +27,7 @@
     "configuration.maven.terminal.customEnv.environmentVariable": "Name of the environment variable to set.",
     "configuration.maven.terminal.customEnv.value": "Value of the environment variable to set.",
     "configuration.maven.view": "Specifies the way of viewing Maven projects.",
+    "configuration.maven.showInExplorerContextMenu": "If this value is true, add a command to create Maven Projects in the Explorer context menu for folders.",
     "configuration.maven.terminal.favorites": "Specify pre-defined favorite commands to execute.",
     "configuration.maven.terminal.favorites.alias": "A short name for the command.",
     "configuration.maven.terminal.favorites.command": "Content of the favorite command.",
--- a/src/Settings.ts
+++ b/src/Settings.ts
@@ -9,6 +9,10 @@ export namespace Settings {
         return ret !== undefined ? ret : [];
     }
 
+    export function showInExplorerContextMenu(): boolean {
+        return !!_getMavenSection<boolean>("maven.showInExplorerContextMenu");
+    }
+
     export function viewType(): string | undefined {
         return _getMavenSection("view");
     }


Eskibear:
Thank you for letting us know. The command entry in context menu was meant to help users get started from scratch, or create submodules into a folder. 

Introducing a new settings makes it more complicated, and hard for users to turn it on. A compromising approach would be, to show the entry only when the seleted item is a folder.  It somehow mitigates the issue in your case. 
```
"when": "explorerResourceIsFolder"
```

> explorerResourceIsFolder: True if a folder is selected in the Explorer.

See [when-clause-contexts](https://code.visualstudio.com/docs/getstarted/keybindings#_when-clause-contexts)


As for`Testplan.md`, it is outdated, you can leave it as is, and we'll remove it later.

IAL32:
> Thank you for letting us know. The command entry in context menu was meant to help users get started from scratch, or create submodules into a folder.

I agree that a new user would need some guidance to create a new project, but I think that not having the option at all makes it less user-friendly overall.

> Introducing a new settings makes it more complicated, and hard for users to turn it on. A compromising approach would be, to show the entry only when the seleted item is a folder. It somehow mitigates the issue in your case.
> 
> ```
> "when": "explorerResourceIsFolder"
> ```

I think that an even more compromising approach would be to show the menu item by default, but still let the user decide whether or not to see it on a per-project basis or on in general on the user's settings.

Eskibear:
@tapeinosyne Are you willing update the PR to adopt `"when": "explorerResourceIsFolder"`?

tapeinosyne:
Certainly we should find a solution that can be merged!

> Introducing a new settings makes it more complicated, and hard for users to turn it on. A compromising approach would be, to show the entry only when the seleted item is a folder. It somehow mitigates the issue in your case.

I appreciate your concern for making the extension approachable, but I would argue that merely reducing the scope where the entry is foisted on the menu would not be a good direction. (Not by coincidence #537 stems from much the same consideration: relevance.)

As a compromise, it would be preferrable to retain the setting, but let `true` be the default: that would allow VS Code users to disable the menu entry should they wish to avoid it, while keeping things as simple as possible for new Maven users. (It also wouldn't alter existing behavior for current users.)

IAL32:
> As a compromise, it would be preferrable to retain the setting, but let `true` be the default: that would allow VS Code users to disable the menu entry should they wish to avoid it, while keeping things as simple as possible for new Maven users. (It also wouldn't alter existing behavior for current users.)

I very much agree, and couldn't have said it better myself. I also agree that a better compromise would be to leave the setting to `true` by default, but then still have the possibility to disable it at will.

Having the option, I think, it's better to not having it at all. Maybe also adding the `when` setting could be a nice layer of customization?

Eskibear:
Agree that we:
* add a new setting with `true` as its default value, but suggest to change a more clear name. It somehow unblocks your scenario.
* also add condition `explorerResourceIsFolder` , as it makes no sense to create a project when you right click on a single file.

tapeinosyne:
> change a more clear name

How about `maven.createFromExplorerContextMenu`?

IAL32:
> `maven.showInExplorerContextMenu`

+1

tapeinosyne:
The PR has been updated. The setting is now:
  - enabled by default
  - shown only if the Explorer context is a folder
  - called `maven.showInExplorerContextMenu`

Eskibear (approved):

PR #562, opened 2020-12-03
can navigate to dependency pom

Eskibear:
image

### DIFF
--- a/src/definition/definitionProvider.ts
+++ b/src/definition/definitionProvider.ts
@@ -1,3 +1,6 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
 import * as _ from "lodash";
 import * as path from "path";
 import * as vscode from "vscode";
@@ -27,14 +30,8 @@ class DefinitionProvider implements vscode.DefinitionProvider {
         const artifactIdHint: string | undefined = artifactIdNode && artifactIdNode.text;
         const versionHint: string | undefined = versionNode && versionNode.text;
         if (groupIdHint && artifactIdHint) {
-          let version: string | undefined;
           const mavenProject: MavenProject | undefined = mavenExplorerProvider.getMavenProject(document.uri.fsPath);
-          if (mavenProject) {
-            version = mavenProject.getDependencyVersion(groupIdHint, artifactIdHint);
-          }
-          if (!version) {
-            version = versionHint;
-          }
+          const version: string | undefined = mavenProject?.getDependencyVersion(groupIdHint, artifactIdHint) || versionHint;
           if (version) {
             const pomPath: string = localPomPath(groupIdHint, artifactIdHint, version);
             return new vscode.Location(vscode.Uri.file(pomPath), new vscode.Position(0, 0));
--- a/src/definition/definitionProvider.ts
+++ b/src/definition/definitionProvider.ts
@@ -1,0 +1,54 @@
+import * as _ from "lodash";
+import * as path from "path";
+import * as vscode from "vscode";
+import { mavenExplorerProvider } from "../explorer/mavenExplorerProvider";
+import { MavenProject } from "../explorer/model/MavenProject";
+import { getMavenLocalRepository } from "../utils/contextUtils";
+import { ElementNode, getCurrentNode, XmlTagName } from "../utils/lexerUtils";
+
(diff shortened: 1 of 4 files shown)


src/definition/definitionProvider.ts
```diff
@@ -0,0 +1,54 @@
+import * as _ from "lodash";
```
jdneo, line 1: license

src/definition/definitionProvider.ts
```diff
@@ -0,0 +1,54 @@
...
+        const groupIdNode: ElementNode | undefined = siblingNodes.find(elem => elem.tag === XmlTagName.GroupId);
+        const versionNode: ElementNode | undefined = siblingNodes.find(elem => elem.tag === XmlTagName.Version);
+        const groupIdHint: string | undefined = groupIdNode && groupIdNode.text;
+        const artifactIdHint: string | undefined = artifactIdNode && artifactIdNode.text;
+        const versionHint: string | undefined = versionNode && versionNode.text;
+        if (groupIdHint && artifactIdHint) {
+          let version: string | undefined;
+          const mavenProject: MavenProject | undefined = mavenExplorerProvider.getMavenProject(document.uri.fsPath);
+          if (mavenProject) {
+            version = mavenProject.getDependencyVersion(groupIdHint, artifactIdHint);
```
jdneo, line 33: `version =  mavenProject.getDependencyVersion(groupIdHint, artifactIdHint) || versionHint;`?

Eskibear: what if `mavenProject` is undefined? We still need below lines.

jdneo: I mean

```
if (mavenProject) {
    version = mavenProject.getDependencyVersion(groupIdHint, artifactIdHint) || versionHint;
}

if (version) {
    const pomPath: string = localPomPath(groupIdHint, artifactIdHint, version);
    return new vscode.Location(vscode.Uri.file(pomPath), new vscode.Position(0, 0));
}
```

Eskibear: In your code above, if `mavenProject` is undefined, then `version` is always undefined, we don't have a chance to assign `versionHint` to it.

jdneo: Ok, then it can be even more simpler

`version = mavenProject?.getDependencyVersion(groupIdHint, artifactIdHint) || versionHint;`

Eskibear: syntactic sugar guru 😄

jdneo (approved):

ghokun:
Thanks for this PR! I see it opens file from .m2 folder.
- Opened file is not readonly.
- Cant navigate further from opened file.
- Opened file has a .pom extension so it is opened as a "Plain Text" file with no syntax highlight. (can be easily fixed with a file association setting)

Should I open issues for these things?

Eskibear:
@ghokun Feel free to open separate issues. Below are some information from my knowledge, just FYI and you can include them in related issues.

> Opened file is not readonly.

I don't think VS Code has provided such API to open files in RO mode. Let me know if you have related information.

> Cant navigate further from opened file.

The feature is not enabled for `.pom` files. We can enable that later, but there's some uncovered cases. E.g. when version is not explicitly written, as below 
```xml
<dependency>
  <groupId>foo</groupId>
  <artifactId>bar</artifactId>
</dependency>
```

If it's in `pom.xml` of your project, we can calculate `version` from effective pom, and navigate to correct file.
If it's in `xxx.pom`, then I don't have a simple way to find out the correct `version`.

Same issue occurs if version is specified by property, e.g. `<version>${azure.version}</version>`. If it fails to calculate effective pom, the navigation functionality will be broken.

> Opened file has a .pom extension so it is opened as a "Plain Text" file with no syntax highlight. (can be easily fixed with a file association setting)

I'm not sure whether the syntax highlighting is provided by other extensions or vscode itself. I can investigate whether it's able to register `.pom` as a subtype of `.xml` files, then we probably get it for free.

ghokun:
@Eskibear 

> I don't think VS Code has provided such API to open files in RO mode. Let me know if you have related information.
While using vscode I realized .class files from dependencies open in a readonly editor. I have looked at the vscode-java extension but could not find any clue.

For other things I will try to open issues and help them if I can.

PR #564, opened 2020-12-08
reuse p2c converter from language client library

Eskibear:


package.json
```diff
@@ -594,6 +594,7 @@
     "md5": "^2.2.1",
     "minimatch": "^3.0.4",
     "vscode-extension-telemetry-wrapper": "^0.8.0",
+    "vscode-languageclient": "^6.1.3",
     "vscode-languageserver-protocol": "^3.15.3",
```
jdneo, line 598: Not quite sure though. Is the module `vscode-languageserver-protocol` used?

Eskibear: yes, they match.  both client@6.1.3 and protocol@3.15.3 are latest for LSP 3.15

jdneo (approved):

PR #581, opened 2021-03-05
add "maven.custom.settingFileLocation"

mannixsuo:
Specifies the absolute path of a maven configuration file


mannixsuo:
I think allows the user to specify the setting.conf file location and find maven local repo by that configuraction file is better than hard coded it.

package.json
```diff
@@ -429,6 +429,12 @@
       {
         "title": "Maven for Java",
         "properties": {
+          "maven.custom.settingFileLocation": {
```
Eskibear, line 432: `maven.settingsFile` would be fine

package.json
```diff
@@ -429,6 +429,12 @@
...
+            "type": "string",
+            "default": "",
+            "description": "%configuration.maven.custom.settingFileLocation%",
+            "scope": "machine-overridable"
```
Eskibear, line 436: Whether allow users to specify settings file per workspace/project?

package.nls.json
```diff
@@ -31,5 +31,6 @@
     "configuration.maven.terminal.favorites": "Specify pre-defined favorite commands to execute.",
     "configuration.maven.terminal.favorites.alias": "A short name for the command.",
     "configuration.maven.terminal.favorites.command": "Content of the favorite command.",
-    "configuration.maven.terminal.favorites.debug": "Whether to execute in debug mode."
+    "configuration.maven.terminal.favorites.debug": "Whether to execute in debug mode.",
+    "configuration.maven.custom.settingFileLocation":"Specifies absolute path of your maven configration file"
```
Eskibear, line 35: 1. typo here
2. the content should be exactly same accross languages.

src/Settings.ts
```diff
@@ -78,6 +78,12 @@ export namespace Settings {
         }
     }
 
+    export namespace UserCustomSetting {
+        export function path(): string | undefined {
+            return workspace.getConfiguration("maven.custom").get<string>("settingFileLocation");
```
Eskibear, line 83: by specifying scope to "machine-overridable" above, here you should allow to get workspace level setting instead of global one.

Eskibear (changes requested):

src/Settings.ts
mannixsuo: It will be better to get this config from workspace configuration, if not find then get from global configuration?

src/Settings.ts
```diff
@@ -78,9 +78,13 @@ export namespace Settings {
         }
     }
 
-    export namespace UserCustomSetting {
+    export namespace SettingsFile {
         export function path(): string | undefined {
-            return workspace.getConfiguration("maven.custom").get<string>("settingFileLocation");
+            let config : WorkspaceFolder | undefined ;
+            if (workspace.workspaceFolders !== undefined) {
+                config =  workspace.workspaceFolders[0];
```
Eskibear, line 85: why always fetching the first workspace folder?

Eskibear (changes requested):

src/Settings.ts
mannixsuo: the first entry corresponds to the value of  `rootPath`

Eskibear: What if it's a multi-root workspace?

mannixsuo: Maybe use `workspace.workspaceFile` here would be better. At first I want this value changed when user select a folder that contains this setting, but I don't find how to get the folder that user current selected。

mannixsuo (commented):
ok

src/Settings.ts
```diff
@@ -82,6 +82,16 @@ export namespace Settings {
         }
     }
 
+    export namespace SettingsFile {
+        export function path(): string | undefined {
+            let config: ConfigurationScope | undefined;
+            if (workspace.workspaceFile !== undefined) {
+                config = workspace.workspaceFile;
+            }
+            return workspace.getConfiguration("maven", config).get<string>("settingsFile");
```
Eskibear, line 91: > workspaceFile
> 		 * Depending on the workspace that is opened, the value will be:
>		 *  * `undefined` when no workspace or  a single folder is opened
>		 *  * the path of the workspace file as `Uri` otherwise. if the workspace
>		 * is untitled, the returned URI will use the `untitled:` scheme
>		 *

You failed at least two scenarios:
1. for an untitled workspace, `config` would be using untitled scheme, and I don't know if it can correctly fetch **workspace** level settings.
2. for multi-root workspace, it's unable to fetch settings in **workspace folder** level.

src/utils/contextUtils.ts
```diff
@@ -28,9 +28,12 @@ export async function loadPackageInfo(context: ExtensionContext): Promise<void>
 
     TEMP_FOLDER_PER_USER = path.join(os.tmpdir(), `${EXTENSION_NAME}-${os.userInfo().username}`);
 
-    // find Maven Local Repositry
+    // find Maven Local Repository
```
Eskibear, line 31: You should add a listener for this setting. Whenever the value is changed, you should also update value of `MAVEN_LOCAL_REPOSITORY`.
I suggest to extract this part into a function, e.g. `loadMavenSettings()`, and re-use it.

src/Settings.ts
```diff
@@ -82,6 +82,16 @@ export namespace Settings {
         }
     }
 
+    export namespace SettingsFile {
```
Eskibear, line 85: Be consistent with the name. Instead of a namespace, simply declare `settingFile` as a function.

Eskibear (changes requested):
> "scope": "machine-overridable"

By declaring it as a **machine**-**overridable** setting, you should take care of workspace/folder level settings when they override the global one.

Another concern is security. Suppose you cloned a repo from GitHub, with malicious workspace settings pointing to "./.m2", with malicious jars also included in this repository under "./.m2/repositories". That could be a problem.

Maven local repository is per machine, while the settings file is actually a global one which can be overrided in project level. I suggest to make it global for simplicity if you cannot find a way to cover all the cases.

src/Settings.ts
mannixsuo: OK I will change it's scope to global .

src/utils/contextUtils.ts
mannixsuo: OK

Eskibear (approved):
Thank you for your contribution!

Issue #590, opened 2021-03-18
Displayed project name missing groupId


yaojinyang:
**Describe the bug**
Displayed project name only uses artifactId defined in .pom. In multi-module project, when modules have same artifactId but different groupIds, Maven view displays same project name, which makes users impossible to distinguish the modules from name. 

**To Reproduce**
1. Import https://github.com/piranhacloud/piranha into VS Code
2. Wait till the whole project is imported 
3. Open Maven view
4. Check project names

**Expected behavior**
Each project should have a unique name.

**Environments (please complete the following information as much as possible):**
 - OS: Windows 10
 - VS Code version: 1.54.3
 - Maven extension version: 0.28.0

**Screenshots**
![Maven issue](https://user-images.githubusercontent.com/54611965/111573066-c279fd80-87e4-11eb-85d8-b93e719600bb.PNG)

mnriem:
Please also consider allowing to switch from the POM groupId:artifactId to the POM name section

Eskibear:
Below change looks fair:
* label of the item: use `name` if provided, `artifactId` as fallback.
* set `groupId:artifactId` as "treeItem.description" which is available now.

![image](https://user-images.githubusercontent.com/2351748/112275551-b39abb80-8cba-11eb-9701-639ed9efe40b.png)

Eskibear:
shipped in v0.29.0

PR #597, opened 2021-03-24
use TreeItem.description to provide show identifier

Eskibear:
Related issue: #590 

Solution:
- Maven items
  - label: show "name" if provided, "artifactId" as fallback.
  - description: always show groupId : artifactId. (version is omitted as I don't think it's that important in distinguishing)
- Plugin items
  - before resolving it to get detailed info (e.g. prefix), show qualified id, aka. groupId:artifactId:version
  - show "prefix" if resolved, and put qualified id in description

After this PR:
image


jdneo:
In the screenshot, there are some description shows as `undefined:xxx`, when will the `undefined` appear?

Eskibear:
Good catch. The case should be when group id is not defined directly in POM, but is inherited from parent POM.

jdneo:
So I guess we need to find some approach to handle this right?

Eskibear:
1. Simplest. If group Id is missing, it's omitted. Because artifact Id must exist, we only show artifact id.
2. Traverse parent POMs above.

Eskibear:
image

@jdneo , please take a look, I adopt option 2) to traverse parents

jdneo (approved):

PR #601, opened 2021-03-26
add lifecycle into maven explorer

Eskibear:
image

the icon "sync" means lifecycle

the icon "gear" is same with IntelliJ IDEA, optimized for developers coming from there.

### DIFF
--- a/src/explorer/model/LifecycleMenu.ts
+++ b/src/explorer/model/LifecycleMenu.ts
@@ -3,7 +3,7 @@
 
 import * as vscode from "vscode";
 import { ITreeItem } from "./ITreeItem";
-import { LifecycleItem } from "./LifecycleItem";
+import { LifecyclePhase } from "./LifecyclePhase";
 import { MavenProject } from "./MavenProject";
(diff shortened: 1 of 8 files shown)


src/explorer/model/LifecycleMenu.ts
```diff
@@ -0,0 +1,25 @@
...
+import { Menu } from "./Menu";
+
+export class LifecycleMenu extends Menu implements ITreeItem {
+    constructor(project: MavenProject) {
+        super(project);
+        this.name = "Lifecycle";
+    }
+
+    public async getChildren() : Promise<LifecycleItem[]> {
+        return ["clean", "validate", "compile", "test", "package", "verify", "install", "site", "deploy"].map(goal => new LifecycleItem(this.project, goal));
```
jdneo, line 17: Should we call them `goal` or `phase`?

Eskibear: ok, renamed to phase

jdneo (approved):

Issue #604, opened 2021-03-31
Monitor status when creating a project from archetype


Eskibear:
To create a project, currently correponding command is assembled and sent to terminal. An disadvantage is, the extension doesn't know whether it's successful or not. 
Ideally if we can monitor the process, we can guide users to open the newly created project, or provide some insightful help against errors.

PR #619, opened 2021-04-22
use vscode.Task to monitor project creation

Eskibear:
This PR inlcudes below changes:
* use vscode.Task to create project , instead of sending command directly to terminal.
* use vscode.QuickInput to pre-input `groupId` and `artifactId`, in order to open correct project folder after creation. (requires refactoring previous multiple-step wizard implementation.)
* embed popular archetypes instead of fetching on-the-fly. (optimize for slow bandwidth/poor networking cases)


src/archetype/createProject/index.ts
```diff
@@ -0,0 +1,49 @@
...
+ * @returns whether all steps are successfully passed
+ */
+export const runSteps = async (steps: IProjectCreationStep[], metadata: IProjectCreationMetadata): Promise<boolean> => {
+  if (!(steps[0] instanceof SelectArchetypeStep)) {
+    throw new Error("First step must be SelectArchetypeStep.");
+  }
+
+  for (let i = 0; i < steps.length; i += 1) {
+    steps[i].nextStep = steps[i + 1];
+    steps[i].previousStep = steps[i - 1];
```
jdneo, line 30: Will index of of bound here?

src/archetype/ArchetypeModule.ts
```diff
@@ -3,104 +3,82 @@
...
             `-DarchetypeArtifactId="${archetypeArtifactId}"`,
             `-DarchetypeGroupId="${archetypeGroupId}"`,
-            `-DarchetypeVersion="${archetypeVersion}"`
+            `-DarchetypeVersion="${archetypeVersion}"`,
+            `-DgroupId="${groupId}"`,
+            `-DartifactId="${artifactId}"`
         ];
         let mvnPath: string | undefined;
-        let cwd: string = targetFolder;
+        let cwd: string | undefined = targetFolder;
```
jdneo, line 67: When cwd === undefined, does that mean user press `esc` when specifying the target folder?

src/archetype/createProject/index.ts
Eskibear: Yes, but doesn't matter.  Suppose n = steps.length, then steps[-1] = steps[n] = undefined.

src/archetype/ArchetypeModule.ts
Eskibear: No. If folder selection is cancelled, nothing will happen, and this method will not be called. Here simply because type definition of `targetFolder` is `string|undefined`. `targetFolder` can not be undefined in real case.

src/archetype/createProject/SpecifyTargetFolderStep.ts
```diff
@@ -0,0 +1,23 @@
...
+
+    public async run(metadata: IProjectCreationMetadata): Promise<StepResult> {
+        // choose target folder.
+        const result: Uri | undefined = await openDialogForFolder({
+            defaultUri: metadata.targetFolder !== undefined ? Uri.file(metadata.targetFolder) : undefined,
+            openLabel: "Select Destination Folder"
+        });
+        const targetFolder: string | undefined = result?.fsPath;
+        if (targetFolder === undefined) {
+            return StepResult.STOP;
```
Eskibear, line 18: @jdneo , regarding to your comments above, if no folder is selected, the series of execution simply stop.

jdneo (approved):

Eskibear:
I'll merge this PR and test against different terminals.

Eskibear:
https://github.com/microsoft/vscode-maven/issues/604

PR #620, opened 2021-04-26
use CMD to create maven project on windows

Eskibear:
if user's default shell is pwsh, args are not correctly passed/escaped


jdneo (approved):

Eskibear:
https://github.com/microsoft/vscode-maven/issues/604

PR #630, opened 2021-05-12
adopt workspace trust API

Eskibear:
See #624 

API is finalized since code v1.56.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -103,9 +103,9 @@
       "dev": true
     },
     "@types/vscode": {
-      "version": "1.43.0",
-      "resolved": "https://registry.npmjs.org/@types/vscode/-/vscode-1.43.0.tgz",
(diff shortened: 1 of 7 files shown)


Eskibear:
#### Maven Wrapper
By default, maven wrapper is preferred if exists. Previously we prompt to ask for user's trust for `mvnw` file found in workspace. Now we look for mvnw files only when workspace is trusted.

#### Options/Envs
By declaring `restrictedConfigurations`, the extension reads workspace level options/envs settings only when workspace is trusted.


https://user-images.githubusercontent.com/2351748/117911598-a724fd80-b310-11eb-997d-84384e62897a.mp4

Eskibear:
Update:

Considering that users should only execute the files you trust, I added a check for all execution-related commands. 

In a restricted workspace:
- view/edit actions should work as before. E.g. POM navigation/editing, adding a dependency.
- for execution related actions, a message box will be prompted guiding you to manage workspace trust.

See a demo below:

https://user-images.githubusercontent.com/2351748/117949964-4532bb00-b345-11eb-86ff-62139b173118.mp4

Eskibear:
close in favor of #631

Issue #559, opened 2020-11-29
Resolve conflicting dependencies


FunnyYish:
## Suggestion

add Dependency Exclusions

## Use Cases

Since Maven resolves dependencies transitively, it is possible for unwanted dependencies to be included in your project's classpath. For example, a certain older jar may have security issues or be incompatible with the Java version you're using. To address this, Maven allows you to exclude specific dependencies. Exclusions are set on a specific dependency in your POM, and are targeted at a specific groupId and artifactId. When you build your project, that artifact will not be added to your project's classpath by way of the dependency in which the exclusion was declared.

Eskibear:
I understand the use case you mentiond. 

> For example, a certain older jar may have security issues or be incompatible with the Java version you're using.

In this case, the first step is to **find out** the incompatible dependency before you perform any action to **exclude** it. After that you take some action to "exclude" it. For the exclusion part, without VS Code, you can always mannually add `<exclusion>`  in pom.xml. So what kind of experience are you expecting? Click dependency node to mark it as "exclusion" or...

FunnyYish:
Maybe like the  idea  [maven-hleper plugin](https://plugins.jetbrains.com/plugin/7179-maven-helper)

Eskibear:
Looks good. As I mentioned above, the pre-requisites of the feature are:
* capability of finding conflicting dependencies. 
  * this extension is written in TS/JS, no convenient library/API for that. 
  * alternative is to parse output of [maven-dependency-plugin](https://maven.apache.org/plugins/maven-dependency-plugin/examples/resolving-conflicts-using-the-dependency-tree.html), but less robust.
* proper UI (vscode doesn't support such a dialog)
  * show dependencies in flat/hierarchical view (see #261)
  * highlight conflicts

Suggestions and PRs are welcome...

Melody618:
hi, @FunnyYish. 

Since 0.32.0, dependency tree has been visualized in maven explorer and dependency exclusions are supported. Here is a [demo.](https://github.com/microsoft/vscode-maven/pull/651#issue-685005501) 
Maybe you can give it a try. Feedbacks, suggestions are welcome.

FunnyYish:
no more better

PR #651, opened 2021-07-07
add "exclude dependency" command to help resolve conflicts

Melody618:
Based on the dependency tree view, add a "exclude dependency" command to resolve the conflicts as mentioned in #559.

https://user-images.githubusercontent.com/25446374/125219572-8e789a00-e2f7-11eb-9a11-3021a4d87735.mp4

Besides, other methods to resolve dependencies conflicts are coming soon.


src/explorer/model/Dependency.ts
```diff
@@ -11,7 +11,12 @@ const CONFLICT_INDICATOR: string = "omitted for conflict";
 export class Dependency extends TreeNode implements ITreeItem {
     private label: string = ""; // groupId:artifactId:version:scope
     private description: string = "";
-    constructor(value: string) {
+    private projectPomPath: string = "";
+    private groupId: string = "";
+    private artifactId: string = "";
+    private version: string = "";
+    private scope: string = "";
+    constructor(value: string, projectPomPath: string) {
```
Eskibear, line 19: better using `constructor(gid, aid, version, scope, projectPom)`. moving parsing logic outside.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -48,3 +50,20 @@ function parseTreeNodes(treecontent: string, separator: string, indent: string,
     }
     return treeNodes;
 }
+
+function setOldestAncestor(treeNodes: Dependency[]): void {
```
Eskibear, line 54: Not necessarily to update `node.root` separately after parsing. You can do it during parsing.

src/utils/lexerUtils.ts
```diff
@@ -7,6 +7,7 @@ export enum XmlTagName {
     ArtifactId = "artifactId",
     Version = "version",
     Dependencies = "dependencies",
+    Exclusions = "exclusions",
```
Eskibear, line 10: In this PR, what's it for?

src/utils/editUtils.ts
```diff
@@ -46,3 +47,11 @@ async function executeRangeFormat(editor: TextEditor, startPosition: Position, l
     editor.selection = new Selection(startPosition, endPosition);
     await commands.executeCommand("editor.action.formatSelection");
 }
+
+export function getIndentation(document: vscode.TextDocument, offset: number): string {
```
Eskibear, line 51: In this PR, why do you extract it here? I don't see any other caller

src/explorer/model/Dependency.ts
```diff
@@ -11,7 +11,12 @@ const CONFLICT_INDICATOR: string = "omitted for conflict";
 export class Dependency extends TreeNode implements ITreeItem {
     private label: string = ""; // groupId:artifactId:version:scope
     private description: string = "";
-    constructor(value: string) {
+    private projectPomPath: string = "";
+    private groupId: string = "";
```
Eskibear, line 15: declared but not used, as reported in Travis error messages.

Eskibear: Do not initialize members as empty string if possible, you can perform null/undefined check later to avoid mistakes.

Eskibear (changes requested):
1. If the PR is not yet ready for review, mark it as a draft PR.
2. Always compile/lint locally to avoid CI failure.

src/utils/editUtils.ts
Melody618: there is another caller in addDependencyHandler.ts. Because I need to use it in excludeDependencyHandler, so I extract the function from addDependencyHandler.ts to editUtils

src/utils/lexerUtils.ts
Melody618: For inserting exclusion node in the pom file

src/handlers/excludeDependencyHandler.ts
```diff
@@ -29,9 +33,9 @@ async function excludeDependency(pomPath: string, gid: string, aid: string, ance
 
     const projectNode: ElementNode = projectNodes[0];
     const dependenciesNode: ElementNode | undefined = projectNode.children && projectNode.children.find(node => node.tag === XmlTagName.Dependencies);
-    const dependencyNode: ElementNode | undefined = dependenciesNode?.children && dependenciesNode?.children.find(node => node.children && node.children[1].tag === XmlTagName.ArtifactId && node.children[1].text === ancestorAid);
+    const dependencyNode: ElementNode | undefined = dependenciesNode?.children && dependenciesNode?.children.find(node => node.children && node.children[1].tag === XmlTagName.ArtifactId && node.children[1].text === rootAid);
     if (dependencyNode === undefined) {
-        throw new UserError("Cannot find the ancestor dependency where to insert Exclusions.");
+        throw new UserError("Cannot find the root dependency where to insert Exclusions.");
```
Eskibear, line 38: "root dependency" doesn't make sense to developers. I suggest to use something like following statement to provide more helpful information.

```
Failed to find the dependency where {gid:aid} is introduced.
```

And it's not a `UserError`, which represents errors caused by malformed settings/input from user.

src/handlers/excludeDependencyHandler.ts
```diff
@@ -8,19 +8,23 @@ import { UserError } from "../utils/errorUtils";
...
         return;
     }
     const pomPath: string = options.projectPomPath;
     if (!await fse.pathExists(pomPath)) {
         throw new UserError("Specified POM file does not exist on file system.");
     }
-    await excludeDependency(pomPath, options.groupId, options.artifactId, options.oldestAncestor.artifactId);
+    let groupId: string;
+    let artifactId: string;
+    [groupId, artifactId] = options.label.split(":");
```
Eskibear, line 21: as you don't modify the values, can be simplified as
```
const [groupId, artifactId] = options.label.split(":");
```

src/handlers/excludeDependencyHandler.ts
```diff
@@ -8,19 +8,23 @@ import { UserError } from "../utils/errorUtils";
 import { ElementNode, getNodesByTag, XmlTagName } from "../utils/lexerUtils";
 
 export async function excludeDependencyHandler(options?: any): Promise<void> {
```
Eskibear, line 10: If options is undefined, there's NPE. 

As an `export` function called by other modules, its signature should be clearly defined instead of optional options of `any` type. If exists, `options` here represents how you want to perform this action.

According to its name, think about what is necessary if you want to exclude a dependency. For me, they are likely to be 1) target pom file path. 2) information of the target dependency to exclude like artifact id and group id.

Or simply `excludeDependencyHandler(toExclude?: Dependency)`.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
     const separator: string = "\r\n";
     const starter: string = "+- ";
-    const treeNodes: Dependency[] = parseTreeNodes(treeContent, separator, indent, starter, eol, project.pomPath);
-    setOldestAncestor(treeNodes);
-    return treeNodes;
+    return parseTreeNodes(treeContent, separator, indent, starter, project.pomPath);
 }
 
-function parseTreeNodes(treecontent: string, separator: string, indent: string, starter: string, eol: string, projectPomPath: string): Dependency[] {
+function parseTreeNodes(treecontent: string, separator: string, indent: string, starter: string, projectPomPath: string): Dependency[] {
```
Eskibear, line 29: `starter` should be `prefix`?

Eskibear: `separator` should be `lineEnding` or `eol`?

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
-            let curNode: Dependency;
-            if (treeChild.indexOf(eol) === -1) {
-                curNode = new Dependency(treeChild, projectPomPath);
+        let curNode: Dependency;
+        let preNode: Dependency;
+        let parentNode: Dependency;
+        let rootNode: Dependency;
+        let curIndex: number;
+        let preIndex: number = 1;
+        const treeChildren: string[] = treecontent.split(separator).splice(1); // delete first line
```
Eskibear, line 38: > // delete first line

everyone knows you are deleting the first line (if you rename separator to eol). Key point is why are you doing this.

Eskibear: rename `treeChildren` to `lines`

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
+        treeChildren.forEach(treeChild => {
```
Eskibear, line 39: `treeChild` is one line of raw string type. You are parsing each line of the content, `lines.forEach(line => ...` would make more sense.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
-function parseTreeNodes(treecontent: string, separator: string, indent: string, starter: string, eol: string, projectPomPath: string): Dependency[] {
+function parseTreeNodes(treecontent: string, separator: string, indent: string, starter: string, projectPomPath: string): Dependency[] {
     const treeNodes: Dependency[] = [];
     if (treecontent) {
-        const treeChildren: string[] = treecontent.split(`${separator}${starter}`).splice(1); // delete first line
-        const toTreeNode = (treeChild: string): Dependency => {
-            let curNode: Dependency;
-            if (treeChild.indexOf(eol) === -1) {
-                curNode = new Dependency(treeChild, projectPomPath);
+        let curNode: Dependency;
```
Eskibear, line 32: To be honest about these variables, I cannot tell the usage from their names.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
+        let preNode: Dependency;
+        let parentNode: Dependency;
+        let rootNode: Dependency;
+        let curIndex: number;
+        let preIndex: number = 1;
+        const treeChildren: string[] = treecontent.split(separator).splice(1); // delete first line
+        treeChildren.forEach(treeChild => {
+            curIndex = treeChild.indexOf(starter);
+            const label: string = treeChild.slice(curIndex + starter.length);
```
Eskibear, line 41: what's `label` of a dependency? For me, label is property of a vscode.TreeItem. For example "full qualified name" of a Maven artifact is `<groupId>:<artifactId>:<version>`, you can later use it as label of a TreeItem, but here `label` doesn't make any sense.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
-        const toTreeNode = (treeChild: string): Dependency => {
-            let curNode: Dependency;
-            if (treeChild.indexOf(eol) === -1) {
-                curNode = new Dependency(treeChild, projectPomPath);
+        let curNode: Dependency;
+        let preNode: Dependency;
+        let parentNode: Dependency;
+        let rootNode: Dependency;
+        let curIndex: number;
+        let preIndex: number = 1;
```
Eskibear, line 37: what's the magic number `1` for?

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -20,50 +20,51 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
+                if (curIndex === preIndex) {
+                    parentNode.addChild(curNode);
+                } else if (curIndex > preIndex) {
+                    parentNode = preNode;
+                    parentNode.addChild(curNode);
                 } else {
-                    node.children.forEach(child => child.oldestAncestor = node.parent?.oldestAncestor);
+                    const level: number = (preIndex - curIndex) / indent.length;
+                    for (let i = level; i > 0; i -= 1) {
+                        parentNode = <Dependency> parentNode.parent;
```
Eskibear, line 57: why do you use type casting here?

Eskibear (changes requested):

src/handlers/excludeDependencyHandler.ts
Melody618: use `vscode.window.showErrorMessage` instead now.

Eskibear: Can `throw new Error(msg)` work? I assume it would also call `vscode.window.showErrorMessage` eventually?

src/explorer/model/Dependency.ts
```diff
@@ -9,24 +10,44 @@ const DUPLICATE_INDICATOR: string = "omitted for duplicate";
...
+    private version: string;
+    private scope: string;
+    private supplement: string = "";
+    constructor(gid: string, aid: string, version: string, scope: string, supplement: string, projectPomPath: string) {
+        super();
+        this.gid = gid;
+        this.aid = aid;
+        this.version = version;
+        this.scope = scope;
+        this.fullArtifactName = join([gid, aid, version, scope], ":");
```
Eskibear, line 26: native `[gid, aid, version, scope].join(":")` should be enough?

src/explorer/model/Dependency.ts
```diff
@@ -9,24 +10,44 @@ const DUPLICATE_INDICATOR: string = "omitted for duplicate";
...
+    }
+
+    public get groupId(): string {
+        return this.gid;
+    }
+
+    public get artifactId(): string {
+        return this.aid;
+    }
+    public get Version(): string {
```
Eskibear, line 45: version

src/explorer/model/Dependency.ts
```diff
@@ -9,24 +10,44 @@ const DUPLICATE_INDICATOR: string = "omitted for duplicate";
...
+        return this.version;
+    }
+    public get Scope(): string {
```
Eskibear, line 48: scope

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -21,48 +21,58 @@ export async function parseRawDependencyDataHandler(project: MavenProject): Prom
...
+                name = name.substr(0, indexCut);
+            }
+            const [gid, aid, version, scope] = name.split(":");
+            return new Dependency(gid, aid, version, scope, supplement, projectPomPath);
+        };
+        lines.forEach(line => {
+            curIndentCnt = line.indexOf(prefix);
+            curNode = toDependency(line);
+            if (curIndentCnt === 0) {
                 curNode.root = curNode;
```
Eskibear, line 54: These nested if-else conditions are a little bit wired. I still believe there's still room to make this logic more concise and clear. But never mind, let's keep it as is, and improve it 
 later.

src/handlers/excludeDependencyHandler.ts
```diff
@@ -3,24 +3,25 @@
...
+    if (toExclude === undefined) {
+        throw new UserError("Only Dependency can be excluded.");
+    }
+    const root: Dependency | undefined =  toExclude.root ? <Dependency> toExclude.root : undefined;
+    if (root === undefined || toExclude.fullName === root.fullName) {
         vscode.window.showInformationMessage("The dependency written in pom can not be excluded.");
         return;
     }
-    const pomPath: string = options.projectPomPath;
+    const pomPath: string = toExclude.ProjectPomPath;
```
Eskibear, line 20: `projectPomPath`? Is it passing typescript compilation?

Eskibear (changes requested):

Eskibear (approved):
Thanks!

PR #656, opened 2021-07-15
add inline entry: add a dependency

Melody618:
In maven explorer, add an inline entry for the "add a dependency" command on the Dependencies node.

https://user-images.githubusercontent.com/25446374/125761755-91e2b4a7-9c7f-4a53-a08a-4cffc04a2140.mp4


src/handlers/addDependencyHandler.ts
```diff
@@ -19,6 +19,9 @@ export async function addDependencyHandler(options?: any): Promise<void> {
     } else if (options && options.projectBasePath) {
         // for "Maven dependencies" nodes from Project Manager
         pomPath = path.join(options.projectBasePath, "pom.xml");
+    } else if (options && options.project.pomPath) {
```
Eskibear, line 22: potential NPE in `options.project.pomPath`.

src/handlers/addDependencyHandler.ts
```diff
@@ -19,6 +19,9 @@ export async function addDependencyHandler(options?: any): Promise<void> {
     } else if (options && options.projectBasePath) {
         // for "Maven dependencies" nodes from Project Manager
         pomPath = path.join(options.projectBasePath, "pom.xml");
+    } else if (options && options.project.pomPath) {
+        // for "Dependencies" node from module in Maven explorer
+        pomPath = options.project.pomPath;
     } else {
         // select a project(pomfile)
         const selectedProject: MavenProject | undefined = await selectProjectIfNecessary();
```
Eskibear, line 27: when options = {}, now you don't enter this branch.

src/handlers/addDependencyHandler.ts
Eskibear: else if (options?.project?.pomPath)

Eskibear (changes requested):

src/handlers/addDependencyHandler.ts
Melody618: when options = {}, you mean when the `DependenciesMenu` is empty? It needs further processed because now the `DependenciesMenu` is hidden if it is empty.

Eskibear: That's a concrete example for NPE mentioned above. `addDependencyHandler` can be called elsewhere, e.g. from Project Manager, or directly from command palette etc.

Eskibear (approved):

PR #659, opened 2021-07-16
add "set dependency version" command

Melody618:
Need further fix and bugs are tracked in #663

Except "exclude the dependency" command in #651, this PR adds another command to help resolve conflicts. It supports setting the dependency version to both those already in the project and those from Maven Central Repository.

https://user-images.githubusercontent.com/25446374/125914271-111767c4-5d58-43ca-ba1c-8bf93e8896e4.mp4


src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+    const fullText: string = project.fullText;
+    const re = new RegExp(`${gid}:${aid}:[\\w.-]+`, "gm");
+    const artifacts: string[] | null = fullText.match(re);
+    let versions: string[] = [];
+    if (artifacts !== null) {
+        artifacts.forEach(a => { versions.push(a.slice(gid.length + aid.length + 2)); });
+    }
+
+    function compare(v1: string, v2: string): number {
```
Eskibear, line 171: Are you sorting them in reverse order?

What's going on with below case?
`["0.1.0", "0.2.0", "0.3.0", "0.10.0"]`

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        if (v1 < v2) {
+            return 1;
+        } else if (v1 > v2) {
+            return -1;
+        } else {
+            return 0;
+        }
+    }
+
+    versions = Array.from(new Set(versions.sort(compare)));
```
Eskibear, line 181: if there's duplicated items, you shoud always first dedup before sorting, for performance concern.

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    await vscode.workspace.applyEdit(edit);
+    const endingPosition: vscode.Position = currentDocument.positionAt(currentDocument.offsetAt(insertPosition) + targetText.length);
+    textEditor.revealRange(new vscode.Range(insertPosition, endingPosition));
+    vscode.workspace.saveAll();
+}
+
+function getAllVersionsInTree(pomPath: string, gid: string, aid: string): string[] {
+    const project: MavenProject | undefined = mavenExplorerProvider.getMavenProject(pomPath);
+    if (project === undefined) {
+        throw new UserError("Failed to get maven projects.");
```
Eskibear, line 161: UserError -> Error

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        const end = dependencyNodeInManagement.contentEnd + dependencyString.length + 1;
+        const range = new vscode.Range(currentDocument.positionAt(start), currentDocument.positionAt(end));
+        edit.delete(currentDocument.uri, range);
+    }
+
+    edit.insert(currentDocument.uri, insertPosition, targetText);
+    await vscode.workspace.applyEdit(edit);
+    const endingPosition: vscode.Position = currentDocument.positionAt(currentDocument.offsetAt(insertPosition) + targetText.length);
+    textEditor.revealRange(new vscode.Range(insertPosition, endingPosition));
+    vscode.workspace.saveAll();
```
Eskibear, line 155: `vscode.workspace.saveAll();` is aggressive. What if you have other unsaved files that you don't want to save for the moment?

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+import { Dependency } from "../explorer/model/Dependency";
+import { MavenProject } from "../explorer/model/MavenProject";
+import { constructDependenciesNode, constructDependencyNode, getIndentation } from "../utils/editUtils";
+import { UserError } from "../utils/errorUtils";
+import { ElementNode, getNodesByTag, XmlTagName } from "../utils/lexerUtils";
+import { getVersions } from "../utils/requestUtils";
+
+const CONFLICT_INDICATOR: string = "omitted for conflict";
+
+export async function setDependencyVersionHandler(selectedItem?: Dependency): Promise<void> {
```
Eskibear, line 16: According to what you defined in `package.json` and `extension.ts`, this handler will only be called upon Dependency items. Why do you allow parameter to be undefined by adding `?` ?

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    if (selectedItem === undefined) {
```
Eskibear, line 17: no need for the null check if you don't allow it.

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        throw new UserError("No dependency node specified.");
+    }
+    const pomPath: string = selectedItem.projectPomPath;
+    if (!await fse.pathExists(pomPath)) {
+        throw new UserError("Specified POM file does not exist on file system.");
+    }
+
+    let effectiveVersion: string;
+    if (selectedItem.supplement.indexOf(CONFLICT_INDICATOR) !== -1) {
```
Eskibear, line 26: Better do this in parsing stage.  Replace `supplement` with an optional data structure, e.g.: 
omitted?: {
reason: "conflict" | "duplicate";
effectiveVersion: string | undefined;
}

You can search gloablly whether you have multiple definition of `CONFLICT_INDICATOR` and calls of `indexOf(CONFLICT_INDICATOR)`?

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    const versions: string[] = getAllVersionsInTree(pomPath, gid, aid);
+    const searchCommand: string = "Search versions from Maven Central Repository...";
+    versions.push(searchCommand);
+
+    const selectedVersion: string | undefined = await vscode.window.showQuickPick(
+        versions.map(version => ({ value: version, label: version !== searchCommand ? `$(package) ${version}` : version, description: version === effectiveVersion ? "effective" : ""})),
+        { placeHolder: `Select a version for ${gid}:${aid}...`,
+            ignoreFocusOut: true}
+    ).then(version => version ? version.value : undefined);
+    if (!selectedVersion) {
```
Eskibear, line 44: explicit use `selected === undefined`

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        const re = /\(omitted for conflict with ([\w.-]+)\)/gm;
+        effectiveVersion = selectedItem.supplement.replace(re, "$1");
+    } else {
+        effectiveVersion = selectedItem.version;
+    }
+
+    const gid: string = selectedItem.groupId;
+    const aid: string = selectedItem.artifactId;
+    const versions: string[] = getAllVersionsInTree(pomPath, gid, aid);
+    const searchCommand: string = "Search versions from Maven Central Repository...";
```
Eskibear, line 36: rename: `OPTION_SEARCH_MAVEN_CENTRAL`

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        versions.map(version => ({ value: version, label: version !== searchCommand ? `$(package) ${version}` : version, description: version === effectiveVersion ? "effective" : ""})),
+        { placeHolder: `Select a version for ${gid}:${aid}...`,
+            ignoreFocusOut: true}
+    ).then(version => version ? version.value : undefined);
+    if (!selectedVersion) {
+        return;
+    }
+    if (selectedVersion === searchCommand) {
+        const selectedVersionFromOrigin: string | undefined = await vscode.window.showQuickPick<vscode.QuickPickItem & { value: string }>(
+            getVersions(gid, aid).then(artifacts => artifacts.map(artifact => ({value: artifact.v, label: `$(package) ${artifact.v}`, description: artifact.v === effectiveVersion ? "effective" : "" }))),
```
Eskibear, line 49: `artifact.v === effectiveVersion ? "effective" : undefined`

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+    const selectedVersion: string | undefined = await vscode.window.showQuickPick(
+        versions.map(version => ({ value: version, label: version !== searchCommand ? `$(package) ${version}` : version, description: version === effectiveVersion ? "effective" : ""})),
+        { placeHolder: `Select a version for ${gid}:${aid}...`,
+            ignoreFocusOut: true}
+    ).then(version => version ? version.value : undefined);
+    if (!selectedVersion) {
+        return;
+    }
+    if (selectedVersion === searchCommand) {
+        const selectedVersionFromOrigin: string | undefined = await vscode.window.showQuickPick<vscode.QuickPickItem & { value: string }>(
```
Eskibear, line 48: Origin -> MavenCentral

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+            { placeHolder: `Select a version for ${gid}:${aid}...`,
+                ignoreFocusOut: true }
+        ).then(artifact => artifact ? artifact.value : undefined);
+        if (!selectedVersionFromOrigin) {
+            return;
+        }
+        if (selectedVersionFromOrigin !== effectiveVersion) {
+            await setDependencyVersion(pomPath, gid, aid, selectedVersionFromOrigin);
+        }
+    } else {
```
Eskibear, line 59: nested if-else loop with multipe place applying changes. The logic can be simplified as below.
```
let selectedVersion;  // sv for short
sv = first quickpick
if (sv === undefined) return;
if (sv === OPTION_MAVEN_CENTRAL)   sv = second quick pick from maven central items
if (sv !== effectiveVersion)  applyChanges


```

src/handlers/setDependencyVersionHandler.ts
Eskibear: try textEditor.document.save

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -0,0 +1,198 @@
...
+        if (selectedVersion !== effectiveVersion) {
+            await setDependencyVersion(pomPath, gid, aid, selectedVersion);
+        }
+    }
+}
+
+async function setDependencyVersion(pomPath: string, gid: string, aid: string, version: string): Promise<void> {
```
Eskibear, line 66: no time to look deep into this part for the moment.

Eskibear (changes requested):
Note: for the pomfile-operation part, I haven't looked deep into it, please test against common cases.

src/handlers/setDependencyVersionHandler.ts
Melody618: It is not necessary here but is for future use such as registering codeActions and the `selectedItem` would be diagnotics objects. So I will just leave it as it is.

src/handlers/setDependencyVersionHandler.ts
```diff
@@ -169,30 +157,30 @@ function getAllVersionsInTree(pomPath: string, gid: string, aid: string): string
...
-            return 0;
+        const indexCut1: number = v1.indexOf("-"); // filter char
+        const indexCut2: number = v2.indexOf("-");
+        if (indexCut1 !== -1) {
+            v1 = v1.substr(0, indexCut1);
+        }
+        if (indexCut2 !== -1) {
+            v2 = v2.substr(0, indexCut2);
+        }
+        const numbers1: number[] = v1.split(".").map(Number);
```
Eskibear, line 168: what if for snapshot versions? E.g. compare `0.1.0-SNAPSHOT` , `0.1.1` and `0.1.2-SNAPSHOT`. 

A systematical way is using `SemVer` lib to compare semantic versions if applied. 

Please create a new issue to track this bug if you don't want to fix it in this PR.

src/explorer/model/OmittedStatus.ts
```diff
@@ -0,0 +1,8 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+export interface IOmittedStatus {
+    status: string;
```
Eskibear, line 5: what if `status` is wrongly set to an arbitary string rather than one of "conflict" "duplicate" "normal" ?

src/explorer/model/Dependency.ts
```diff
@@ -4,31 +4,29 @@
...
+    private _omittedStatus: IOmittedStatus;
+    constructor(gid: string, aid: string, version: string, scope: string, omittedStatus: IOmittedStatus, projectPomPath: string) {
         super();
         this._gid = gid;
         this._aid = aid;
         this._version = version;
         this._scope = scope;
         this.fullArtifactName = [gid, aid, version, scope].join(":");
-        this._supplement = supplement;
+        this._omittedStatus = omittedStatus;
```
Eskibear, line 25: It's a little bit strange to use omittedStatus.status=="normal" to represent a normal dependency. 

Imagine this normal dependency has a property `omittedStatus`, and the status is "normal". So literally is it omitted?

package.nls.json
```diff
@@ -1,40 +1,41 @@
-{
```
Eskibear, line 1: why is the whole file changed?

Eskibear (changes requested):
Some nits. And if you want to get this merged to unblock other PRs, and to fix the issues in future refactoring, please let me know.

src/explorer/model/Dependency.ts
Melody618: I think it's ok. `omittedStatus` is more like `omittedOrNot`, maybe I can rename it in a later fix.

Eskibear: yes, it presents "omittedOrNot", that's why I was suggesting an optional property like `omitted?:IOmittedStatus`.

Eskibear: never mind, refactor it later.

Issue #261, opened 2019-02-26
Better visualization of dependency tree


Eskibear:
It's convenient to for conflict resolving. 

Now I have to execute "mvn dependency:tree -Doutput..." to print the tree to a file, and then open it in VSCode for further analysis. 
I think a better visualization with conflicts highlighted would be very helpful.

Eskibear:
My first thought is to use a "WebView" UI, open in another column

Eskibear:
Now it's only plain text generated by command `dependency:tree`. It's better to define a new file schema, and add intellisense for that, e.g. highlighting conflicts.

mukteshkrmishra:
Any updates on this?

Eskibear:
Not yet, I don't have bandwidth for it recently. BTW, discussion, suggestions and PRs are welcome.

mukteshkrmishra:
I can work on this.

Eskibear:
@mukteshkrmishra Are you able to show dependencies successfully by right clicking the project? I just find a bug that breaks the feature all the time...

mukteshkrmishra:
Not yet successful.

Eskibear:
Got it. And it has been fixed in [#398](https://github.com/microsoft/vscode-maven/pull/398/files#diff-fac897be847e65eb786e16505b0a9711R37-R38) and will be included in the next release.

mukteshkrmishra:
Awesome @Eskibear . Thank You

brunovieira97:
Any progress on this? I noted the menu for this in VS Code removes vertical tree lines (pipes) too...

Eskibear:
Since 0.32.0, dependency tree has been visualized in maven explorer.

Melody618:
hi, @mukteshkrmishra @brunovieira97 

Since 0.32.0, the dependency tree has been visualized in maven explorer. And it looks like below.

![Snipaste_2021-08-18_14-22-36](https://user-images.githubusercontent.com/25446374/129848036-b7770d72-8817-4f10-904b-ff6a86fc8cc6.png)


If you are still interested in dependency visualization, maybe you can give it a try. Feedbacks, suggestions, and feature requests are welcome.

PR #664, opened 2021-07-23
add conflictMessage in Dependency and implement fileDecoration

Melody618:
This pr adds a new parameter in class `Dependency` to track conflicts for both fileDecoration and diagnostics.

Also, it colors the items that have conflicts to make them more conspicuous.

part of #261 

image

### DIFF
--- a/src/explorer/decorationProvider.ts
+++ b/src/explorer/decorationProvider.ts
@@ -2,11 +2,10 @@
 // Licensed under the MIT license.
 
 import * as vscode from "vscode";
-import { Disposable } from "vscode";
 
 class DecorationProvider implements vscode.FileDecorationProvider {
-    private disposables: Disposable[] = [];
-    private decoration: vscode.FileDecoration = new vscode.FileDecoration("", "has conflicts", new vscode.ThemeColor("inputValidation.warningBorder"));
+    private disposables: vscode.Disposable[] = [];
+    private decoration: vscode.FileDecoration = {color: new vscode.ThemeColor("list.warningForeground")};
     constructor() {
         this.decoration.propagate = true;
         this.disposables.push(vscode.window.registerFileDecorationProvider(this));
--- a/src/explorer/decorationProvider.ts
+++ b/src/explorer/decorationProvider.ts
@@ -7,7 +7,6 @@ class DecorationProvider implements vscode.FileDecorationProvider {
     private disposables: vscode.Disposable[] = [];
     private decoration: vscode.FileDecoration = {color: new vscode.ThemeColor("list.warningForeground")};
     constructor() {
-        this.decoration.propagate = true;
         this.disposables.push(vscode.window.registerFileDecorationProvider(this));
     }
 
--- a/src/explorer/decorationProvider.ts
+++ b/src/explorer/decorationProvider.ts
@@ -12,13 +12,13 @@ class DecorationProvider implements vscode.FileDecorationProvider {
     }
 
     public provideFileDecoration(uri: vscode.Uri): vscode.ProviderResult<vscode.FileDecoration> {
-        const endWithMessage: string | undefined = uri.toString().split("/").pop();
-        if (endWithMessage === "hasConflict") {
+        if (uri.query === "hasConflict") {
             return this.decoration;
         } else {
             return null;
         }
     }
+
     public dispose(): void {
         this.disposables.forEach(d => d.dispose());
     }
--- a/src/explorer/decorationProvider.ts
+++ b/src/explorer/decorationProvider.ts
@@ -1,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { Disposable } from "vscode";
+
+class DecorationProvider implements vscode.FileDecorationProvider {
+    private disposables: Disposable[];
+    private decoration: vscode.FileDecoration = new vscode.FileDecoration("", "has conflicts", new vscode.ThemeColor("inputValidation.warningBorder"));
+    constructor() {
+        this.disposables = [];
+        this.decoration.propagate = true;
+        this.disposables.push(vscode.window.registerFileDecorationProvider(this));
+    }
+
+    public provideFileDecoration(uri: vscode.Uri): vscode.ProviderResult<vscode.FileDecoration> {
+        const endWithMessage: string | undefined = uri.toString().split("/").pop();
+        if (endWithMessage === "hasConflict") {
+            return this.decoration;
+        } else {
+            return null;
+        }
+    }
+    public dispose(): void {
+        this.disposables.forEach(d => d.dispose());
+    }
+}
+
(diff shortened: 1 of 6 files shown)


src/explorer/decorationProvider.ts
```diff
@@ -0,0 +1,29 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { Disposable } from "vscode";
```
Eskibear, line 5: keep consistency. don't have to import it.

src/explorer/decorationProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+
+class DecorationProvider implements vscode.FileDecorationProvider {
+    private disposables: Disposable[];
```
Eskibear: vscode.Disposable

src/explorer/decorationProvider.ts
```diff
@@ -0,0 +1,29 @@
...
+    private decoration: vscode.FileDecoration = new vscode.FileDecoration("", "has conflicts", new vscode.ThemeColor("inputValidation.warningBorder"));
```
Eskibear, line 9: use undefined instead of empty string.  Don't use empty string unless you have a clear reason.

Eskibear: `inputValidation.warningBorder` doesn't make sense to me. it's not in a input box, and why use border color? 

> list.warningForeground: Foreground color of list items containing warnings.

maybe this one?

src/explorer/decorationProvider.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+import * as vscode from "vscode";
+import { Disposable } from "vscode";
+
+class DecorationProvider implements vscode.FileDecorationProvider {
+    private disposables: Disposable[] = [];
```
Eskibear, line 8: vscode.Disposible

src/explorer/model/Dependency.ts
```diff
@@ -15,6 +16,7 @@ export class Dependency extends TreeNode implements ITreeItem {
     private _version: string;
     private _scope: string;
     private _omittedStatus: IOmittedStatus;
+    public conflictMessages: IConflictMessage[] = [];
```
Eskibear, line 19: I don't know why you need this. gid/aid are props. version1/version2 can be retrieved from omittedStatus.

src/explorer/model/Dependency.ts
```diff
@@ -50,6 +56,16 @@ export class Dependency extends TreeNode implements ITreeItem {
         return this._scope;
     }
 
+    public get uri(): vscode.Uri {
+        const filePath: string = localPomPath(this._gid, this._aid, this._version);
+        let uri: vscode.Uri = vscode.Uri.file(filePath);
+        uri = uri.with({authority: this.projectPomPath});
```
Eskibear, line 62: why update `authority`?

Eskibear (changes requested):

src/explorer/model/Dependency.ts
Melody618: I use it to track children's `conflictMessages`.

src/explorer/model/Dependency.ts
Melody618: In multi-module projects, one dependency with the same `localPomPath` may import in different Maven modules. In order to distinguish them (maybe one has conflicts but the other one does not), I use `authority`.

src/explorer/decorationProvider.ts
Melody618: Thanks! I just search for `warning` and choose the first color.

src/explorer/model/Dependency.ts
Eskibear: you should take notes for any tricks (anti common sense) in comments.

src/explorer/model/Dependency.ts
Eskibear:
```
        /**
         * A flag expressing that this decoration should be
         * propagated to its parents.
         */
        propagate?: boolean;
```

you've already propagate the decoration from bottom up. why do you need children's conflict info?

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -84,6 +89,25 @@ function parseTreeNodes(treecontent: string, eol: string, indent: string, prefix
...
+            uri = uri.with({authority: projectPomPath}); // distinguish dependency in multi-module project
+            if (curNode.omittedStatus.status === "conflict") {
+                curNode.uri = uri.with({query: "hasConflict"});
+                // find all parent and set hasConflict upforward
+                let tmpNode = curNode;
+                while (tmpNode.parent !== undefined) {
+                    const parent = <Dependency> tmpNode.parent;
+                    if (parent.uri.query !== "hasConflict") {
+                        parent.uri = uri.with({query: "hasConflict"});
+                    }
```
Eskibear, line 106: Pruning: I assume if the parsing is top-down, here you can add `else break;`.  Correct me if I'm wrong.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -66,10 +68,13 @@ function parseTreeNodes(treecontent: string, eol: string, indent: string, prefix
         lines.forEach(line => {
             curIndentCnt = line.indexOf(prefix);
             curNode = toDependency(line);
+            let uri: vscode.Uri;
+            let curFilePath: string;
             if (curIndentCnt === 0) {
                 curNode.root = curNode;
                 rootNode = curNode;
                 parentNode = curNode;
+                curFilePath = `${curNode.groupId}.${curNode.artifactId}`;
```
Eskibear, line 77: Can it be `groupid/artifactid`? That would make more sense. Or is there any pitfall?

Eskibear (changes requested):
some nits.

src/handlers/parseRawDependencyDataHandler.ts
Melody618: since now `propagate = false`,  `groupid/artifactid` is ok. Otherwise, the same `groupid` may import errors.

src/handlers/parseRawDependencyDataHandler.ts
```diff
@@ -74,7 +74,7 @@ function parseTreeNodes(treecontent: string, eol: string, indent: string, prefix
                 curNode.root = curNode;
                 rootNode = curNode;
                 parentNode = curNode;
-                curFilePath = `${curNode.groupId}.${curNode.artifactId}`;
+                curFilePath = path.join(`${curNode.groupId}`, `${curNode.artifactId}`);
```
Eskibear, line 77: unnecessary string templates.

`${var}` = `var`

Eskibear (approved):

jdneo:
Sorry for the late call, is there anything that blocks us using grey color to represent omitted state?

PR #675, opened 2021-07-27
refractor code action

Melody618:
add `map<diagnostic, Dependency>` in `diagnosticProvider`, so when generating code action, we can get `Dependency` for quick fix.


Eskibear (approved):
LGTM.

PR #680, opened 2021-07-28
use content provider to open epom and dependencies

Eskibear:
Previously opened as an untitled file, which has below disadvantages:
- unreadable title
- prompt for saving on closing file

Now use content provider for better UX.

### DIFF
--- a/src/contentProvider.ts
+++ b/src/contentProvider.ts
@@ -10,7 +10,7 @@ import { Utils } from "./utils/Utils";
 
 class MavenContentProvider implements vscode.TextDocumentContentProvider {
     public async provideTextDocumentContent(uri: vscode.Uri, _token: vscode.CancellationToken): Promise<string | undefined> {
-        if (uri.scheme !== "maven") {
+        if (uri.scheme !== "vscode-maven") {
             throw new Error(`Scheme ${uri.scheme} not supported by this content provider.`);
         }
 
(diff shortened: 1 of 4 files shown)


Eskibear:
Before:
image
image


After:
image

src/contentProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+                    return Utils.getEffectivePom(pomPath);
+                }
+            default:
+        }
+        return undefined;
+
+    }
+}
+
+export const contentProvider: MavenContentProvider = new MavenContentProvider();
```
jdneo, line 36: add a new line

src/contentProvider.ts
```diff
@@ -0,0 +1,36 @@
...
+import * as vscode from "vscode";
+import { mavenExplorerProvider } from "./explorer/mavenExplorerProvider";
+import { IEffectivePom } from "./explorer/model/IEffectivePom";
+import { MavenProject } from "./explorer/model/MavenProject";
+import { getDependencyTree } from "./handlers/showDependenciesHandler";
+import { Utils } from "./utils/Utils";
+
+class MavenContentProvider implements vscode.TextDocumentContentProvider {
+    public async provideTextDocumentContent(uri: vscode.Uri, _token: vscode.CancellationToken): Promise<string | undefined> {
+        if (uri.scheme !== "maven") {
```
jdneo, line 13: Consider use scheme like `vscode-maven` to specify this is a vscode context based scheme?

Just like the VS Code team does for the virtual file system, which is called `vscode-vfs`.

jdneo (approved):

PR #683, opened 2021-07-28
use SemVer lib to compare versions

### LINKED ISSUE #669: Use `SemVer` lib to compare versions

Melody618:
fix #669. use `SemVer` lib to compare dependency versions. For those who do not follow SemVer policy, correct them and then compare.

### DIFF
--- a/src/handlers/setDependencyVersionHandler.ts
+++ b/src/handlers/setDependencyVersionHandler.ts
@@ -2,6 +2,7 @@
 // Licensed under the MIT license.
 
 import * as fse from "fs-extra";
+import * as semver from "semver";
 import * as vscode from "vscode";
 import { mavenExplorerProvider } from "../explorer/mavenExplorerProvider";
 import { Dependency } from "../explorer/model/Dependency";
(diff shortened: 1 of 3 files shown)


src/handlers/setDependencyVersionHandler.ts
```diff
@@ -164,28 +165,12 @@ function getAllVersionsInTree(pomPath: string, gid: string, aid: string): string
...
-                i += 1;
-            }
-        }
-        return 0;
+        // correct versions that do not follow SemVer Policy
+        const s1: semver.SemVer | null = semver.coerce(v1);
+        const s2: semver.SemVer | null = semver.coerce(v2);
+        const version1: semver.SemVer | string = s1 === null ? v1 : s1;
+        const version2: semver.SemVer | string = s2 === null ? v2 : s2;
+        return semver.rcompare(version1, version2);
```
Eskibear, line 173: What's the result if one of the versions is not a strict semver? what's the difference if we add `loose` here?

Eskibear (approved):

Issue #615, opened 2021-04-16
Can't format the dependency after execute 2~3 times the command: maven: add dependency


SunnyChen1:
**To Reproduce**
1. Open pom.xml 
2. F1 to execute maven: add dependency(e.g. : function)
3. Repeat step2 to 2~3 times
4. See dependency not be format

**Expected behavior**
 Dependency should be format automatically

**Environments:**
 - OS: Windows 10
 - VS Code version: 1.55.0
 - Extension version 0.29.0&Old version
**Screenshots**
![image](https://user-images.githubusercontent.com/63633836/114993218-2cfa8800-9ece-11eb-8ed6-6dd00b9e5a4e.png)

Eskibear:
known issue, because of applying textEdit on dirty files, so the position is wrong. Possible mitigation can be either: 
1. we save the document before applying it.
2. read content from textDocument instead of file (preferred).

Eskibear:
Closing as completed. #690 fixed dependency insertion against dirty documents, which was the diagnosed root cause.

PR #690, opened 2021-08-03
fix: insert malformed depenedency when pom.xml is dirty

Eskibear:
known issue #615


Eskibear:
@Melody618 please also apply this change to "exclude dependency" and "set dependency version" features.

jdneo (approved):

Melody618 (approved):

PR #695, opened 2021-08-09
fix: diagnostics do not refresh after editing pom file

### LINKED ISSUE #692: diagnostics not refresh after editing pom.xml
**Describe the bug**
error message in Problems channel for conflict dependencies has the wrong location after editing pom.xml

**Expected behavior**
refresh diagnostics when the pom.xml has been changed

**Screenshots**
image

Melody618:
fix #692, refresh diagnostics if the pom file has changed.

Also apply adaptive debounce to improve the performance.

### DIFF
--- a/src/DiagnosticProvider.ts
+++ b/src/DiagnosticProvider.ts
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT license.
+import * as _ from "lodash";
 import * as vscode from "vscode";
 import { Dependency } from "./explorer/model/Dependency";
 import { UserError } from "./utils/errorUtils";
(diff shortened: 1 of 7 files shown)


Melody618:
still working on adaptive debounce

jdneo:
Debouncing part looks good to me, while I didn't quite get when `conflictNodes` needs to be a member of `DiagnosticProvider`

Eskibear:
@jdneo please take a look at the moving avg impl.

jdneo:
As I said, the debouncing part looks good to me. But I'm not sure persisting `conflictNodes` inside `diagnosticProvider` is a good practice. 

Does that mean all the `conflictNodes` belongs to the same uri?

Melody618:
> Does that mean all the conflictNodes belongs to the same uri?

Yes, one pom file, one uri, and the `conflictNodes` hold all conflict dependencies in this pom file.

Melody618:
now `DiagnosticProvider` is workspace level and it provides diagnostics for all pom files. However, it will hold only conflict nodes for one pom file and it is a module level.

I am working to refine the structure: remove `conflictNodes`  and add a utility function to link uri and conflict nodes.

src/DiagnosticProvider.ts
```diff
@@ -16,22 +20,44 @@ class DiagnosticProvider {
...
+        context.subscriptions.push(vscode.workspace.onDidChangeTextDocument(
+        async e => {
+            if (e.document.fileName.endsWith("pom.xml")) {
+                await this.debouncedRefresh(e.document.uri);
+            }
         }));
     }
 
-    public async refreshDiagnostics(uri: vscode.Uri, conflictNodes: Dependency[]): Promise<vscode.Diagnostic[]> {
-        this.map.clear();
```
Melody618, line 26: should not clear map because refresh diagnostics for one pom file should not change diagnostics of other pom files.

src/DiagnosticProvider.ts
```diff
@@ -44,9 +45,14 @@ class DiagnosticProvider {
     }
 
     public async refreshDiagnostics(uri: vscode.Uri): Promise<void> {
-        this.map.clear();
         const diagnostics: vscode.Diagnostic[] = [];
-        for (const node of this.conflictNodes) {
+        const project: MavenProject | undefined = mavenExplorerProvider.getMavenProject(uri.fsPath);
+        if (project === undefined) {
+            throw new Error("Failed to get maven projects.");
```
jdneo, line 51: plural should not be used here. A uri should only map to one project, right?

jdneo (approved):

PR #699, opened 2021-08-11
refine right-click command entry for dependencies

Melody618:
this pr makes three changes:

1. root dependencies do not have exclude command
2. only conflict dependencies have set-version command
3. change wording "set version to" to "resolve conflict"

### DIFF
--- a/package.json
+++ b/package.json
@@ -454,7 +454,7 @@
         },
         {
           "command": "maven.project.setDependencyVersion",
-          "when": "view == mavenProjects && viewItem == Dependency",
+          "when": "view == mavenProjects && viewItem =~ /\\w*Dependency/",
           "group": "0-dependency@1"
         },
         {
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -449,12 +449,12 @@
         },
         {
           "command": "maven.project.excludeDependency",
-          "when": "view == mavenProjects && viewItem == Dependency",
+          "when": "viewItem == Dependency || viewItem == conflictDependency && view == mavenProjects",
```
Eskibear, line 452: For `||` and `&&`, which operator has higher priority?

Melody618: oops, wrong understanding but get right results.
I have tried `"view == mavenProjects && viewItem =~ /(conflict){0,1}Dependency/"` but it doesn't work. What is the correct way to represent `"Dependency || conflictDependency"` ?

Melody618: now use `view == mavenProjects && viewItem =~ /Dependency | conflictDependency/` and it works.

Eskibear (approved):

Issue #701, opened 2021-08-11
Feature request: Double click on lifecycle phase can start the phase


doggy8088:
## Suggestion

I'd like to suggestion "double-click" on the lifecycle can trigger the lifecycle phase to start.

## Use Cases

To start an lifecycle phase, the ▶ icon is too small. I have to move the mouse precisely to trigger the phase. This is not convenient for some people.

![image](https://user-images.githubusercontent.com/88981/129062792-d3b2baaa-3330-4c7e-b283-ebdef114e4ce.png)

## Examples

See `Use Cases`.

Eskibear:
As far as I know, there's no "double click" handler on tree item. (let me know if you find it)

one option is to add an entry in context menu, so you can `Right click -> Run` just like:
![image](https://user-images.githubusercontent.com/2351748/129171217-679ed197-d115-4cb0-ab77-1532b5d11bea.png)

doggy8088:
I think `Right click -> Run` is an acceptable option. Thanks.

Eskibear:
FYI `Right click -> Run` is available in v0.33.0.

doggy8088:
@Eskibear I saw it. Thanks!

But the "Chinese" display the Simplified Chinese. Do you have a Traditional Chinese translation?

![image](https://user-images.githubusercontent.com/88981/131617375-75e604fb-9516-4512-b9b3-11a848dddc9a.png)

Eskibear:
Currently we only have a single `package.nls.zh.json` for Chinese localization. You can create a PR contributing a `package.nls.zh-tw.json` if you are willing to do that.

doggy8088:
@Eskibear 
I saw two terms not been translated in `package.nls.zh.json`. Just let you know. 

1. contributes.viewsWelcome.mavenProjects.untrustedWorkspaces
2. configuration.maven.showInExplorerContextMenu

PR #702, opened 2021-08-12
Lifecycle phase item: add run entry in context menu

Eskibear:
see https://github.com/microsoft/vscode-maven/issues/701#issuecomment-897478010


jdneo:
Looks good but curious whether vs code support double click action in tree view?

Eskibear:
AFAIK, they don't. Do you know the related issue requesting double-click API?

Eskibear:
I don't find any double click API, but only an old issue. https://github.com/microsoft/vscode/issues/85636

jdneo (approved):

PR #704, opened 2021-08-18
fix goToDefinition: refine contextValue and when-clause

Melody618:
this pr fixes the wrong entries of `goToDefinition` and refine context value and when-clause for future use.

format contextValue: `Dependency[+<type>]`
possible type: conflict, root

### DIFF
--- a/package.json
+++ b/package.json
@@ -449,17 +449,17 @@
         },
         {
           "command": "maven.project.excludeDependency",
-          "when": "view == mavenProjects && viewItem =~ /^(?!.*?\\+root)/",
+          "when": "view == mavenProjects && viewItem =~ /maven:dependency(?!.*?\\b\\+root\\b)/",
           "group": "0-dependency@0"
         },
         {
           "command": "maven.project.setDependencyVersion",
-          "when": "view == mavenProjects && viewItem =~ /Dependency(?=.*?\\b\\+conflict\\b)/",
+          "when": "view == mavenProjects && viewItem =~ /maven:dependency(?=.*?\\b\\+conflict\\b)/",
           "group": "0-dependency@1"
         },
         {
           "command": "maven.project.goToDefinition",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -449,17 +449,17 @@
         },
         {
           "command": "maven.project.excludeDependency",
-          "when": "view == mavenProjects && viewItem =~ /^(Dependency)|(conflictDependency)$/",
+          "when": "view == mavenProjects && viewItem =~ /^(?!.*?\\+root)/",
```
Eskibear, line 452: will the command show on other non-dependency node?

Eskibear (changes requested):
let's follow the same convention with https://github.com/microsoft/vscode-java-dependency/wiki/Register-Command-onto-the-Nodes-of-Project-View 
Then you can use the same regexp for when clauses.

i.e. 
`maven:dependency+conflict`

package.json
Eskibear: @jdneo 
Here we have these node types: `maven:dependency[+root/+conflict]`. Is there a regexp to query "all dependency nodes with no +root property"? (I assume there should be a negative lookahead mechanism)

@Melody618  If the answer is no, we can also workaround it, by adding `+nonRoot` to non-root nodes.

jdneo: Try `maven:dependency(?=.*?\b\+conflict\b)(?!.*?\b\+root\b)`

Eskibear: image

Cool.

@Melody618 please use above regexp.
@jdneo suggest to add `(?!.*?\b\+exclude_prop\b)` in Wiki.

Melody618: as the picture shows, the regexp is not able to hit `maven:dependency`

Eskibear: of course it's as expected. for your case, simply append `(?!.*?\b\+root\b)`.

Eskibear (approved):

PR #715, opened 2021-08-22
Add maven.warnConflictsInDiagnostics configuration setting

### LINKED ISSUE #677: Can dismiss the warnings in the pom.xml
Project which has several dependencies will have such conflicts with a high possibility. It could be better if some controls can give to the user, such as:

- code action to dismiss the warnings in a pom.xml
- turn off the conflicts detection (Maybe he just want to see all the direct dependencies and their versions)

Melody618:
fix #677


package.json
```diff
@@ -667,6 +667,12 @@
             "default": false,
             "description": "%configuration.maven.pomfile.prefetchEffectivePom%",
             "scope": "application"
+          },
+          "maven.warnConflictsInDiagnostics": {
```
Eskibear, line 671: I searched "diagnostics" in existing settings, TS is using 
 "tsserver.enableProjectDiagnostics". So maybe something like "maven.dependency.enableConflictDiagnostics` would be better? Please comment on name of this setting.  @jdneo

package.nls.json
```diff
@@ -37,5 +37,6 @@
     "configuration.maven.terminal.favorites.alias": "A short name for the command.",
     "configuration.maven.terminal.favorites.command": "Content of the favorite command.",
     "configuration.maven.terminal.favorites.debug": "Whether to execute in debug mode.",
-    "configuration.maven.settingsFile": "Specifies the absolute path of your maven configuration file, the default value is ~/.m2/settings.xml"
+    "configuration.maven.settingsFile": "Specifies the absolute path of your maven configuration file, the default value is ~/.m2/settings.xml",
+    "configuration.maven.warnConflictsInDiagnostics": "Specify whether to show conflict dependencies in pom.xml files or Problems channel."
```
Eskibear, line 41: Better to refine the wording.

> in pom.xml files or Problems channel.

"Problems channel" is not accurate, and I think the point is to emphasize:
1) whether to show corresponding dianostics.
2) diagnotics are about dependency conflict.

src/Settings.ts
```diff
@@ -13,6 +13,10 @@ export namespace Settings {
         return !!_getMavenSection<boolean>("maven.showInExplorerContextMenu");
     }
 
+    export function warnConflictsInDiagnostics(): boolean {
+        return !!_getMavenSection<boolean>("warnConflictsInDiagnostics");
```
Eskibear, line 17: For a "resource" scope setting, you should get its value passing a "resource uri". Or to make it "window"-scoped

package.json
jdneo: +1 for maven.dependency.enableConflictDiagnostics

Eskibear: @melody618 let's use this name

Eskibear (approved):

Issue #662, opened 2021-07-23
Automatically open the new created maven project


jdneo:
reference: https://www.reddit.com/r/vscode/comments/o8muvl/maven_for_java_extension_automatically_open/

Or make a setting to control it.

Anyway, if a user creates a new Maven project, it looks like having a large possibility to open it to do sth after that.

Eskibear:
>  it looks like having a large possibility to open it 

Yes. But if a single project is created, I don't think it's that annoying to ask for confirmation only once. 

Technically it's feasible to add a setting `maven.project.defaultOpenMethod`, and set it to empty by default.  


BTW, in Spring Initializr we have this setting as required by 3rd party integration.

Eskibear:
Do you want me to include it in the coming release?

PR #717, opened 2021-08-23
add setting for project opening method

Eskibear:
for #662

### DIFF
--- a/package.json
+++ b/package.json
@@ -673,6 +673,17 @@
             "default": "true",
             "description": "%configuration.maven.dependency.enableConflictDiagnostics%",
             "scope": "window"
+          },
+          "maven.defaultOpenProjectMethod": {
+            "default": "Interactive",
+            "type": "string",
+            "scope": "window",
+            "description": "%configuration.maven.defaultOpenProjectMethod%",
+            "enum": [
+              "Interactive",
+              "Open",
+              "Add to Workspace"
+            ]
           }
         }
       }
--- a/package.json
+++ b/package.json
@@ -674,11 +674,11 @@
             "description": "%configuration.maven.dependency.enableConflictDiagnostics%",
             "scope": "window"
           },
-          "maven.defaultOpenProjectMethod": {
+          "maven.projectOpenBehavior": {
             "default": "Interactive",
             "type": "string",
             "scope": "window",
-            "description": "%configuration.maven.defaultOpenProjectMethod%",
+            "description": "%configuration.maven.projectOpenBehavior%",
             "enum": [
               "Interactive",
               "Open",
--- a/package.nls.json
+++ b/package.nls.json
@@ -38,5 +38,6 @@
     "configuration.maven.terminal.favorites.command": "Content of the favorite command.",
     "configuration.maven.terminal.favorites.debug": "Whether to execute in debug mode.",
     "configuration.maven.settingsFile": "Specifies the absolute path of your maven configuration file, the default value is ~/.m2/settings.xml",
-    "configuration.maven.dependency.enableConflictDiagnostics": "Specify whether to show diagnostics for conflict dependencies."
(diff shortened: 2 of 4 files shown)


jdneo:
Ci fails

package.json
```diff
@@ -673,6 +673,17 @@
             "default": "true",
             "description": "%configuration.maven.dependency.enableConflictDiagnostics%",
             "scope": "window"
+          },
+          "maven.defaultOpenProjectMethod": {
```
jdneo, line 677: maven.projectOpenBehavior

package.json
```diff
@@ -673,6 +673,17 @@
...
+            "default": "Interactive",
```
jdneo, line 678: Do we have data for the favorite option? I prefer `Open` personally

package.nls.json
```diff
@@ -38,5 +38,6 @@
     "configuration.maven.terminal.favorites.command": "Content of the favorite command.",
     "configuration.maven.terminal.favorites.debug": "Whether to execute in debug mode.",
     "configuration.maven.settingsFile": "Specifies the absolute path of your maven configuration file, the default value is ~/.m2/settings.xml",
-    "configuration.maven.dependency.enableConflictDiagnostics": "Specify whether to show diagnostics for conflict dependencies."
+    "configuration.maven.dependency.enableConflictDiagnostics": "Specify whether to show diagnostics for conflict dependencies.",
+    "configuration.maven.defaultOpenProjectMethod": "Default method of opening newly created project."
```
jdneo, line 42: Default behavior xxx...

src/extension.ts
```diff
@@ -256,26 +256,32 @@ async function openPomHandler(node: MavenProject | { uri: string }): Promise<voi
 }
 
 function registerProjectCreationEndListener(context: vscode.ExtensionContext): void {
+    const OPEN_IN_NEW_WORKSPACE = "Open";
+    const OPEN_IN_CURRENT_WORKSPACE = "Add to Workspace";
+
+    const specifyOpenMethod = async (hasOpenFolder: boolean, projectName: string, projectLocation: string) => {
+        let openMethod = vscode.workspace.getConfiguration("maven").get<string>("defaultOpenProjectMethod");
```
jdneo, line 263: openMethod -> openBehavior

Remember to update other occurrences as well.

package.json
Eskibear: I directly followed spring-initializr, see https://github.com/microsoft/vscode-spring-initializr/pull/157

A better name in my mind was `maven.project.postCreationBehavior"

package.json
Eskibear: > Do we have data for the favorite option?

No we don't. We can add corresponding telemetry. Then we need to keep "interactive" as default for a while to collect data.

I'm 50:50, maybe prefer interactive a little bit. maybe my key point is, I don't think this notification is a frequent thing.

package.json
jdneo: I'm referring from https://github.com/microsoft/vscode-azurefunctions/blob/54c7614fbb984fbf3ff6bff1028979bd96f3a644/package.json#L952

postCreation sounds like a lifecycle method name but not a setting name

package.json
jdneo: Colleting data sounds good

package.json
Eskibear: then let's follow azfunc

package.json
Eskibear: will sendInfo

Eskibear:
telemetry schema: 

```
info {
  name: proejectOpenBehavior(from setting/choice)
  value: open/add-to-workspace/cancelled
}
```
@jdneo see if it works

jdneo (approved):

Eskibear:
pending event to tag:   `info { value: string }`

PR #722, opened 2021-08-30
add refresh to DependenciesMenu

Melody618:
add `refresh()` for DependenciesMenu to save dirty pom file.

### DIFF
--- a/src/explorer/model/DependenciesMenu.ts
+++ b/src/explorer/model/DependenciesMenu.ts
@@ -54,7 +54,7 @@ export class DependenciesMenu extends Menu implements ITreeItem {
 
     private _savePom(): void {
         const textEditors: vscode.TextEditor[] = vscode.window.visibleTextEditors;
-        const textEditor: vscode.TextEditor | undefined = textEditors.find(editor => editor.document.uri === vscode.Uri.file(this.project.pomPath));
+        const textEditor: vscode.TextEditor | undefined = textEditors.find(editor => editor.document.fileName === this.project.pomPath);
         if (textEditor !== undefined) {
             textEditor.document.save();
         }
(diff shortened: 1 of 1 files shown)


src/explorer/model/DependenciesMenu.ts
```diff
@@ -45,4 +46,17 @@ export class DependenciesMenu extends Menu implements ITreeItem {
...
     }
+
+    public refresh(): void {
+        this._savePom();
+        mavenExplorerProvider.refresh(this);
+    }
+
+    private _savePom(): void {
+        const textEditors: vscode.TextEditor[] = vscode.window.visibleTextEditors;
+        const textEditor: vscode.TextEditor | undefined = textEditors.find(editor => editor.document.fileName === this.project.pomPath);
```
Eskibear, line 57: It would be better to use `uri` for future compatibility (when it's not fs-based). Can you give it a try?

something like `document.uri === vscode.uri.file(pomPath)`

Eskibear (commented):
LGTM, better to adopt URI-based comparision.

src/explorer/model/DependenciesMenu.ts
```diff
@@ -54,7 +54,8 @@ export class DependenciesMenu extends Menu implements ITreeItem {
 
     private _savePom(): void {
         const textEditors: vscode.TextEditor[] = vscode.window.visibleTextEditors;
-        const textEditor: vscode.TextEditor | undefined = textEditors.find(editor => editor.document.fileName === this.project.pomPath);
+        const textEditor: vscode.TextEditor | undefined = textEditors.find(editor =>
```
Eskibear, line 57: ` vscode.Uri.file(this.project.pomPath)` might be expensive. It's better to construct it outside.

Eskibear (approved):
LGTM

PR #723, opened 2021-09-01
Enable CodeQL in Azure Pipelines

Eskibear:


.azure-pipelines/ci.yml
```diff
@@ -17,7 +17,20 @@ jobs:
...
-
+  - job: CodeQL
+    displayName: Code QL
+    pool:
+      vmImage: 'windows-latest'
+    steps:
+      - task: Semmle@1
+        inputs:
+          sourceCodeDirectory: '$(Build.SourcesDirectory)'
+          language: 'tsandjs'
```
jdneo, line 28: Will you enable it on Java?

Eskibear: good question, let me try if I can enable java together.

Eskibear: seems we cannot set multiple languages for the same task, so I added another task for java. hope it works.

Eskibear:
Now I move all static analysis tasks into one job, including credscan and CodeQL for ts/js/java.

jdneo (approved):

PR #725, opened 2021-09-01
improve project open behavior

Eskibear:
This PR:
- Add "None" option to `maven.projectOpenBehavior`.
- Update settings section in README


package.json
```diff
@@ -696,7 +696,8 @@
             "enum": [
               "Interactive",
               "Open",
-              "Add to Workspace"
+              "Add to Workspace",
+              "None"
```
jdneo, line 700: Is this value referenced in code base?

Eskibear: No. 

In `specifyOpenMethod`
If value of setting is "Interactive", we prompt to ask.
Otherwise value of setting is directly returned.

Based on above result:
If Open ...
If Add To Workspace ...

For the case `None`, we simply do nothing.

jdneo (approved):

Issue #658, opened 2021-07-16
can not automatically add a dependency whose type is pom


Melody618:
**Describe the bug**
can not automatically add a dependency whose type is not jar, the default value.

**To Reproduce**
1. Run `add a dependency` command
2. Search for `takari-plugin-integration-testing` and add, which is a pom dependency
3. See error in pom.xml

**Expected behavior**
support other types of dependencies

**Environments:**
 - OS: Windows 10
 - VS Code version: 1.58.1
 - Extension version: 0.31.0

**Screenshots**
![image](https://user-images.githubusercontent.com/25446374/125902046-da3cbec3-5ad2-4743-8f0e-cc77d6bdc008.png)

Melody618:
both `addDependency` and `setDependencyVersion` only support dependencies whose type is jar. So we need to fix them separately. 
1. for `addDependency`: get `type` from `getArtifacts()`,  and change the construction of dependencies node in pom.xml
2. for `setDependencyVersion`: add `-DshowTypes` in maven command, add `type` in class `Dependency` and change related functions.

PR #726, opened 2021-09-01
fix add a dependency: add support for all types dependencies

Melody618:
fix part of #658

now users can add a dependency with all kinds of types, including jar, pom, war, etc.


src/handlers/setDependencyVersionHandler.ts
```diff
@@ -108,10 +108,11 @@ async function insertDependencyManagement(pomPath: string, targetNode: ElementNo
             node.children?.find(id => id.tag === XmlTagName.GroupId && id.text === gid) &&
             node.children?.find(id => id.tag === XmlTagName.ArtifactId && id.text === aid)
         );
-        targetText = constructDependencyNode(gid, aid, version, `${baseIndent}${indent}`, indent, eol);
+        const newIndent: string = `${baseIndent}${indent}`;
+        targetText = constructDependencyNode({gid: gid, aid: aid, version: version, baseIndent: newIndent, indent: indent, eol: eol});
     } else if (targetNode.tag === XmlTagName.DependencyManagement && dependenciesNode === undefined) {
         insertPosition = currentDocument.positionAt(targetNode.contentStart);
-        targetText = constructDependenciesNode(gid, aid, version, baseIndent, indent, eol);
+        targetText = constructDependenciesNode({gid: gid, aid: aid, version: version, baseIndent: baseIndent, indent: indent, eol: eol});
```
Eskibear, line 115: Can be simplified to {gid, aid, ...}.   because {gid} is short for {gid: gid} if the key name is same as variable.

Eskibear (commented):
LGTM with some nits.

PR #733, opened 2021-09-13
fix maven executable path for PowerShell

Eskibear:
for earlier version of maven, `mvn.bat` is the executable instead of `mvn.cmd`.


Eskibear:
See https://github.com/microsoft/vscode-maven/issues/489#issuecomment-917613597

jdneo (approved):

PR #734, opened 2021-09-13
open .pom file in RO mode

### LINKED ISSUE #718: open POM under local maven repo in read-only mode
It's a follow-up item of #428 

So far vscode has no such API, but we can workaround it via ContentProvider API, the same way we show effectivePOM and dependency list now.

Eskibear:
fix #718 

image

### DIFF
--- a/src/contentProvider.ts
+++ b/src/contentProvider.ts
@@ -45,10 +45,12 @@ class MavenContentProvider implements vscode.TextDocumentContentProvider {
                 } else {
                     return Utils.getEffectivePom(pomPath);
                 }
-            default:
-                const fsUri = uri.with({scheme: "file"});
+            case "local-repository":
+                const fsUri = uri.with({ scheme: "file", authority: "" });
                 return (await vscode.workspace.fs.readFile(fsUri)).toString();
+            default:
         }
+        return undefined;
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/contentProvider.ts
```diff
@@ -8,6 +8,12 @@ import { MavenProject } from "./explorer/model/MavenProject";
 import { getDependencyTree } from "./handlers/showDependenciesHandler";
 import { Utils } from "./utils/Utils";
 
+/**
+ * URI patterns.
+ * vscode-maven://dependencies/<pom-path>/Dependencies?<pom-path>
+ * vscode-maven://effective-pom/<pom-path>/EffectivePOM.xml?<pom-path>
+ * vscode-maven:///<pom-path-in-local-maven-repository>
```
jdneo, line 15: three `/`s here, as expected?

CsCherrYY, line 15: how about using a specific `authority` like the previous ones?

Eskibear: as expected. Here authority is empty, so `//<empty>` is authority. And `path` starts with "/".

> URI = scheme:[//authority]path[?query][#fragment]

I use `authority` to differentiate dependencies/effective-pom/common pom file.

Eskibear: yes, it's an option.

Eskibear: Let's call it "local-repository"? What if later we want to open a project pom.xml in RO mode?

CsCherrYY (approved):

src/contentProvider.ts
Eskibear: > What if later we want to open a project pom.xml in RO mode?

no such requirement for the moment. let's rename / use another authority if later we need  do that.

Issue #735, opened 2021-09-14
Action to add Maven Wrapper


brunoborges:
## Suggestion

Provide an action that adds the Maven Wrapper to an existing Maven project.

Eskibear:
A command "Add Maven Wrapper" should work when there's no wrapper detected in project root. Source of the wrapper can be:

#### Option 1:
If a local `mvn` is found on the machine, we run `mvn -N io.takari:maven:0.7.7:wrapper`. Otherwise we copy [maven wrapper packed in this extension](https://github.com/microsoft/vscode-maven/tree/main/resources/maven-wrapper) to project folder.

Pros: 
We allow to specify maven version ` -Dmaven=<version>`.  (but do they care much about maven version?)

#### Option 2:
Always copy the packed wrapper into project folder.

Pros:
Simple and straightforward, we don't require a local maven. (that's why some developers need wrapper).

brunoborges:
As far as I understand, VS Code Maven Extension already comes with a Maven bundled within. This Maven installation could be used instead, both as a Fallback for when there is no Maven Runtime configured, and as a runtime to execute the Add Maven Wrapper goal.

Eskibear:
> VS Code Maven Extension already comes with a Maven bundled within.

No. To clarify, a Maven **wrapper** (instead of a Maven distribution) is bundled in this extension. That was due to some consideration:
- ~10MB extra size 
- unclear whether there's legal issue

brunoborges:
Got it! It is still possible to use the Maven Wrapper to fall the Maven Wrapper plugin to add the Maven Wrapper to a project :-)

Eskibear:
BTW, here is a quick demo of this feature, triggered from command palatte. I'm thinking about whether we need more entries.

https://user-images.githubusercontent.com/2351748/133648682-0ffac8c3-42d4-4d97-b2c6-d6ad158a9030.mp4

brunoborges:
Option B is certainly faster. I'd only suggest to allow user input (optionally) for which Maven version they want (latest by default).

Eskibear:
Agree to allow user input of Maven version.

Eskibear:
@HvB To unblock yourself, perhaps you can contribute to the project on your own by submitting a pull request. 

You can start from https://github.com/microsoft/vscode-maven/pull/739 which is one of the simple solutions for this.

PR #739, opened 2021-09-17
support adding maven wrapper

Eskibear:
See #735 

- Support to "Add maven wrapper".
  - currently only triggerred  from command palatte. 
  - it's added to workspace root folder. (we can discuss if it should be added per project in the future).

- BTW I updated tasks.json and build script, as new webpack-cli changes arguments and output format.

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -18,8 +18,8 @@
                 ],
                 "background": {
                     "activeOnStart": true,
-                    "beginsPattern": "Compilation \\w+ starting…",
-                    "endsPattern": "Compilation\\s+finished"
+                    "beginsPattern": "\\w+",
+                    "endsPattern": "webpack .* compiled"
                 }
             },
             "isBackground": true,
--- a/package.json
+++ b/package.json
@@ -224,6 +224,11 @@
         "command": "maven.favorites",
         "title": "%contributes.commands.maven.favorites%",
         "category": "Maven"
+      },
+      {
+        "command": "maven.project.addWrapper",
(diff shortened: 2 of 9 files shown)


Eskibear:
Demo here:

https://user-images.githubusercontent.com/2351748/133741814-59b9ec32-1a7a-49ec-a960-a643e8a642e6.mp4

jdneo:
Haven't go deep into the PR yet. Some quick question:

1. Will users have the need to specify the wrapper's version?
2. The maven plugin has the knowledge for all the pom.xml's location, will the out-most pom be better than the workspace root?

Eskibear:
> 1. Will users have the need to specify the wrapper's version?

I don’t think so, wrapper version doesn’t affect the project, as long as the wrapper is working well. Allowing to specify Maven version should be enough.


> 2. The maven plugin has the knowledge for all the pom.xml's location, will the out-most pom be better than the workspace root?

As far as I remember, VS Code Maven extension picks maven wrapper under workspace root, and that’s why I create wrapper it here. To get best user experience for both Java extension and Maven extension, it’s also recommended to open root project. In that case, Pom.xml is just located at workspace root.

But I’m open to the option that we base on out-most POM, and the target folder should at least contain a Pom.xml

Eskibear:
I'll ask user to choose target location of maven wrapper, candidates will be the location containing pom.xml files. WDYT? @jdneo

jdneo:
Sounds good to me.

Eskibear:
closing obsolete PR.

PR #743, opened 2021-09-23
Enable external extension calls to AddDependencies

dsimpsonSinia:
## Where I started
I am writing an extension to enhance Mulesoft dependency support within vscode, but I didn't want to duplicate all your code if the users will already have the maven extension installed. The problem is when the `maven.project.addDependency` command is called, it opens a prompt for the user rather than allowing the command to be passed the dependency it should add. So, I added an `if..else` to your command handler to just add the dependency if it was passed in through the options.

## What it turned in to

While making these changes, I also need to be able to set a classifier on a dependency, so I reworked the `constructDependencyNode` in editUtils to support that extra field and make the method more condensed. Otherwise there would have been an explosion of permutations for optional fields.

Instead of also modifying the `constructDependenciesNode` method to match the singular form, I just made it reuse the existing method and added an extra bump to the indentation to keep the file clean.

Since I had already made so many changes, I just went ahead and did the same for `constructDependencyManagementNode`.


Eskibear (approved):
LGTM, thanks!

PR #768, opened 2021-12-21
execute mvn with maven.settingsFile

emliunix:
Now that we have maven.settingsFile config, it seems natural to execute mvn with it (which I found out not.)

Is there any concern not doing so.

### DIFF
--- a/src/handlers/addDependencyHandler.ts
+++ b/src/handlers/addDependencyHandler.ts
@@ -35,34 +35,41 @@ export async function addDependencyHandler(options?: any): Promise<void> {
         throw new UserError("Specified POM file does not exist on file system.");
     }
 
-    const keywordString: string | undefined = await vscode.window.showInputBox({
-        ignoreFocusOut: true,
-        prompt: "Input keywords to search artifacts from Maven Central Repository.",
-        placeHolder: "e.g. spring azure storage",
-        validateInput: (text: string) => {
-            if (text.trim().length < 3) {
-                return "Keywords are too short.";
-            }
-            return undefined;
-        }
-    });
-    if (!keywordString) {
-        return;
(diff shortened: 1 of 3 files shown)


Eskibear:
Thank you for contribution. My only concern is, if it works when there's whitespace in `maven.settingsFile` without escaping? Can you test it against some common terminals? e.g. cmd/powershell in Windows.

Eskibear:
BTW, code to create project with `mvn archetype:generate` is in below location, you should also update it there:

https://github.com/microsoft/vscode-maven/blob/1de8c75c38553d66be4438806103475aebe11a3f/src/archetype/ArchetypeModule.ts#L77

Eskibear:
@emliunix We'd like to get this fixed within this month. Do you have time to follow up and update the PR accordingly?

emliunix:
Hi @Eskibear , sorry the PR was accidentally closed by a unclean rebase. I didn't figure out how to re-open it so I re-created a new PR #776.

emliunix:
> Thank you for contribution. My only concern is, if it works when there's whitespace in `maven.settingsFile` without escaping? Can you test it against some common terminals? e.g. cmd/powershell in Windows.

I looked how other code handles a path, seems the effective solution is to simply quote the path in the Args string. Though I think it's not a complete "escaping", it works in bash/cmd/powershell as I tested.

Is there a better/unified utility to escape a path that I can use?

Eskibear:
Quoting is what I'm currently using to add `-f path-to-pom.xml`, though it still fails to cover some corner case. 

> Is there a better/unified utility to escape a path that I can use?

I didn't find a perfect one when I drafted related code. Escaping chars are different for different shells, e.g. `^`  ` """` and so on. Please let me know if you find any.

PR #775, opened 2022-01-07
allow external call to create project with predefined args

Eskibear:
Can pass an `IProjectCreationMetadata` as argument to create Maven projects, which will avoid unnecessary user interactions.

### DIFF
--- a/src/archetype/createProject/SpecifyArchetypeVersionStep.ts
+++ b/src/archetype/createProject/SpecifyArchetypeVersionStep.ts
@@ -8,6 +8,10 @@ export class SpecifyArchetypeVersionStep implements IProjectCreationStep {
     public previousStep?: IProjectCreationStep;
 
     public async run(metadata: IProjectCreationMetadata): Promise<StepResult> {
+        if (metadata.archetypeGroupId && metadata.archetypeArtifactId && metadata.archetypeVersion) {
+            return StepResult.NEXT;
+        }
+
         const disposables: Disposable[] = [];
         const specifyAchetypeVersionPromise = new Promise<StepResult>((resolve, reject) => {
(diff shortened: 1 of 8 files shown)


src/archetype/createProject/SpecifyArchetypeVersionStep.ts
```diff
@@ -8,6 +8,10 @@ export class SpecifyArchetypeVersionStep implements IProjectCreationStep {
     public previousStep?: IProjectCreationStep;
 
     public async run(metadata: IProjectCreationMetadata): Promise<StepResult> {
+        if (metadata.archetypeGroupId && metadata.archetypeArtifactId && metadata.archetypeVersion) {
```
jdneo, line 11: Since this step is to check archetype version, is it necessary to check `archetypeGroupId` & `archetypeArtifactId`?

Eskibear: I agree the check looks strange here. But version is meaningful only when we have groupId and artifactId.

Internal workflow is:
- select archetype from list, so we have groupId and artifactId and a list of valid versions.
- in this step, select a version from the list.

For external calls, if no groupId and artifactId from args:
- Currently it fallbacks to pick a version, but without a list of versions, it throws en error message in Line 18.
- If we don't check these information, it will assemble a malformed command line, and fail in terminal. 

I'm ok with both approaches. I want to know you preference after my explanation.

jdneo: I see. It looks good to me now. I guess we can keep the current check until people from community ask for more customization.

jdneo (approved):

PR #776, opened 2022-01-08
execute mvn with maven.settingsFile

emliunix:
Re-create PR #768 because it's accidentally closed.

Original description:
> Now that we have maven.settingsFile config, it seems natural to execute mvn with it (which I found out not.)
> Is there any concern not doing so.


Eskibear (approved):
LGTM

Eskibear:
@jdneo
To verify:
1. open a Maven project. 
2. modify setting `maven.settingsFile` and point it to a valid settings.xml
3. run a maven goal, verify if `-s settings.xml` is appended to the command.
4. create a new project, also verify command line in the task.

Issue #763, opened 2021-12-03
Maven reload/dependency update fails on StackOverflowException


yntelectual:
**Describe the bug**

After I open a Java project, then resolving maven dependencies fail, VSCode showing message: 
`Background process terminated with code 1.` and in the Output of `Maven for Java` I see a StackOverflowException - attached bellow.

**To Reproduce**
1 open java project
2 expand maven dependencies

**Expected behavior**
Maven dependencies are resolved and shown

**Environments (please complete the following information as much as possible):**
 - OS: Windows 10/WSL2
 - VS Code version:  1.62.2
 - Extension version v0.34.1

**Additional context**
```
[[1;34mINFO[m] [1m---------------< [0;36mde.telekom.tm.ibt:supplier-cockpit-ui[0;1m >----------------[m
[[1;34mINFO[m] [1mBuilding supplier-cockpit-ui 1.0.0-SNAPSHOT[m
[[1;34mINFO[m] [1m--------------------------------[ war ]---------------------------------[m
[[1;33mWARNING[m] The POM for com.sun.xml.bind:jaxb-osgi:jar:2.2.10 is invalid, transitive dependencies (if any) will not be available, enable debug logging for more details
[[1;34mINFO[m] 
[[1;34mINFO[m] [1m--- [0;32mdepgraph-maven-plugin:3.3.0:graph[m [1m(default-cli)[m @ [36msupplier-cockpit-ui[0;1m ---[m
[[1;34mINFO[m] [1m------------------------------------------------------------------------[m
[[1;34mINFO[m] [1;31mBUILD FAILURE[m
[[1;34mINFO[m] [1m------------------------------------------------------------------------[m
[[1;34mINFO[m] Total time:  6.728 s
[[1;34mINFO[m] Finished at: 2021-12-03T10:52:38+01:00
[[1;34mINFO[m] [1m------------------------------------------------------------------------[m
---------------------------------------------------
constituent[0]: file:/usr/share/maven/conf/logging/
constituent[1]: file:/usr/share/maven/lib/plexus-cipher.jar
constituent[2]: file:/usr/share/maven/lib/maven-builder-support-3.x.jar
constituent[3]: file:/usr/share/maven/lib/guice.jar
constituent[4]: file:/usr/share/maven/lib/maven-resolver-spi.jar
constituent[5]: file:/usr/share/maven/lib/maven-resolver-api.jar
constituent[6]: file:/usr/share/maven/lib/commons-lang3.jar
[...]

PR #780, opened 2022-01-12
depgraph-maven-plugin@3.3.1 fixed transitive deps

Eskibear:
https://github.com/microsoft/vscode-maven/issues/763#issuecomment-1008081292


jdneo (approved):

Eskibear:
To verify:
1. open any maven project.
2. add dependency `arquillian-chameleon-junit-container-starter`
3. expand `Dependencies`, expect a success.
4. check pom.xml, it should report problems (circular dependencies).

BTW, previous version would failed with StackOverflowException in output.

PR #796, opened 2022-02-23
fix completion range

### LINKED ISSUE #787: Version autocompletion should replace existing version
## Suggestion

When autocompleting a version, the textedit should replace the existing version. Currently the new version is added to the one you're trying to replace, which forces you to manually delete the old one. Makes a poor UX. 

See:

Jan-31-2022 10-38-11

Eskibear:
close #787 

Same issue occur not only for version, but also group Id and artifact Id.

### DIFF
--- a/src/completion/completionProvider.ts
+++ b/src/completion/completionProvider.ts
@@ -38,7 +38,7 @@ class CompletionProvider implements vscode.CompletionItemProvider {
             return undefined;
         }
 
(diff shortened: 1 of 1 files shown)


Eskibear:
Now: 

https://user-images.githubusercontent.com/2351748/155253454-30264e73-ad6b-4727-8835-ab82cb4ea18b.mp4

jdneo (approved):

PR #825, opened 2022-06-21
initialize extension tests

Eskibear:


Eskibear:
A simple case added for the first time. Result as below.

```
[main 2022-06-21T02:42:16.333Z] update#setState idle

[main 2022-06-21T02:42:17.584Z] Starting extension host with pid 23536 (fork() took 16 ms).

Loading development extension at <redacted>

  Maven Project View Tests

    ✔ Can list maven projects (112ms)

  1 passing (140ms)

[main 2022-06-21T02:42:28.045Z] Waiting for extension host with pid 23536 to exit.

[main 2022-06-21T02:42:28.059Z] Extension host with pid 23536 exited with code: 0, signal: null.

Exit code:   0
Done
```

CsCherrYY:
The test framework `vscode-test` has been renamed to `@vscode/test-electron` and the current version (1.5.1) has been outdated for about one year. It makes us 

1. must add something like `--disable-workspace-trust` in the running argument
2. are not able to run test with an existing VS Code instance.

and so on. Could you update the test framework to the newest version and see if there is any breaking change?

Eskibear:
@CsCherrYY   Updated. So far so good in my case. The only difference is you have to specify a `platform` when downloading target vscode packages.

CsCherrYY:
the changes LGTM. How can we trigger the tests in CI? via Github Actions or Azure pipeline?

Eskibear:
I'm thinking of setting up GitHub Actions later.

CsCherrYY (approved):
Sounds good. you can refer to https://github.com/microsoft/vscode-java-dependency/blob/main/.github/workflows/linux.yml

jdneo (approved):

Issue #843, opened 2022-07-19
Better completion for properties


Eskibear:
E.g. for Spring-boot projects, version of different starters are defined via properties, e.g. `<activemq.version>5.16.5</activemq.version>`, in parent poms. And if you want to use another version, simply change the corresponding properties.

If, vscode-maven extension can suggestion all available properties, that'd be helpful.

Eskibear:
impl in #844
Not that only when you have effective pom generated, the extension will provide properties as candidates.

PR #844, opened 2022-07-19
completion for properties

Eskibear:
see #843

### DIFF
--- a/src/completion/completionProvider.ts
+++ b/src/completion/completionProvider.ts
@@ -16,16 +16,17 @@ const artifactSegments: string[] = [
     "\t<artifactId>$2</artifactId>",
     "\t<version>$3</version>"
 ];
-const dependencySnippetString: string = [
+const dependencySnippetString = (eol: string) => [
     "<dependency>",
     ...artifactSegments,
     "</dependency>"
-].join("\n");
-const pluginSnippetString: string = [
+].join(eol);
+
+const pluginSnippetString = (eol: string) => [
     "<plugin>",
     ...artifactSegments,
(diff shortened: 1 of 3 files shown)


src/completion/completionProvider.ts
```diff
@@ -149,6 +150,23 @@ class CompletionProvider implements vscode.CompletionItemProvider {
...
                 return new vscode.CompletionList([snippetItem], false);
             }
+            case XmlTagName.Properties: {
+                const project = mavenExplorerProvider.getMavenProject(document.uri.fsPath);
+                const props = await project?.getProperties();
+
+                if (props) {
+                    const propertyToCompletionItem = (prop: string) => {
+                        const item = new vscode.CompletionItem(prop, vscode.CompletionItemKind.Property);
+                        const insertText = `<${prop}>$1</${prop}>\n$0`;
```
jdneo, line 160: Will VS Code automatically correct the EOL according to user's document preference?

Eskibear: Yes, I think so, `\n` is used for all the other snippets and so far so good. See line 28 of this file.

It's a good point and later I'll consider whether to use `os.EOL` instead, for all snippets.

jdneo: You can get it from the input `TextDocument`

jdneo: Change the `\n` here as well?

Eskibear: updated. also extracted the util function for re-use.

jdneo (approved):

PR #834, opened 2022-07-11
add setting for format of project node name

Eskibear:
See https://github.com/microsoft/vscode-maven/issues/794#issuecomment-1177524076

Adding a new setting `maven.explorer.projectName` to specify format of project node name.

PR #851, opened 2022-07-25
show correct version when it's inherited from parent

Eskibear:
image


CsCherrYY (approved):

Eskibear:
please verify according to https://github.com/microsoft/vscode-maven/pull/834#issuecomment-1193731419

PR #856, opened 2022-08-01
update dependencies

Eskibear:

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -37,26 +37,26 @@
       "dev": true
     },
     "@jridgewell/gen-mapping": {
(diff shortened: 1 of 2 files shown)


jdneo:
Will you consider updating the [libs](https://github.com/microsoft/vscode-maven/blob/main/jdtls.ext/com.microsoft.java.maven.plugin/.classpath) as well?

Eskibear:
Thank you for reminding me. I'll update target platform too.

Eskibear:
@jdneo
I won't update those `libs`, as there would be a lots of breaking changes.

PR #870, opened 2022-09-08
Created a new maven goal to execute the 'clean install' in just one command

gustavosimon:
Created a new maven goal to execute the 'clean install' in just one command.
It was created to take the same behavior that NetBeans IDE provides with Shift + F11.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the "vscode-maven" extension will be documented in this file.
 
+## 0.39.0
+### Added
+- Created a new maven goal to execute 'clean install' in just one command.
+
 ## 0.38.0
 ### Added
 - Support Maven schema-based completion for pom.xml files. [#857](https://github.com/microsoft/vscode-maven/pull/857)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to the "vscode-maven" extension will be documented in this f
 
 ## 0.39.0
 ### Added
-- Created a new maven goal to execute 'clean install' in just one command.
+- Created a new maven goal to execute 'clean install' in just one command. [#870] (https://github.com/microsoft/vscode-maven/pull/870)
 
 ## 0.38.0
 ### Added
--- a/package.json
+++ b/package.json
@@ -1,802 +1,823 @@
 {
-  "name": "vscode-maven",
-  "displayName": "Maven for Java",
-  "description": "%description%",
-  "version": "0.38.0",
-  "icon": "resources/logo.png",
-  "publisher": "vscjava",
-  "preview": true,
-  "aiKey": "7c4292ac-8ca8-4e02-9f1c-0b73e1eeeca4",
-  "engines": {
-    "vscode": "^1.65.0"
-  },
-  "enabledApiProposals": [],
-  "capabilities": {
-    "virtualWorkspaces": false,
-    "untrustedWorkspaces": {
-      "supported": "limited",
-      "restrictedConfigurations": [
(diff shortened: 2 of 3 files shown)


Eskibear:
Thank you for the contribution. 
I know that "clean install" is a commonly used command. Currently we have several ways for you to easily execute it:
- Right click to run it from `Custom...` entry, and re-run it with command `Maven: History...`
- Set it as a favorite command, and run it with command `Maven: Favorites...`


The biggest concern about this PR is we want to keep context menu clear and concise, instead of being crowded by the all the commands. Particularly, in macOS you have to scroll down to view all items if it's too long. If we added "clean install" into context menu, then what about other commands like "clean compile", "clean package", etc.?

gustavosimon:
Hi @Eskibear !

Thank you for your feedback!!

Studying the extension, I saw that exists the Custom and Favorites commands to do it, but I think that's not productive, because I have to do it several times in a day.

So, this is because I think that is important than the command be in main menu.

I think that is important too because it will help a lot of people that is switching from Netbeans to VSCode, using your extension.

I think that we can add commands like "clean compile", "clean package", etc into the main menu too. 

What do you think about it? 

Thanks again about taking me a feedback! Have a nice day!

Eskibear:
image

Let's clarify the terms first. There're two places holding the goals shortcuts: "Lifecycle" menu (A), context menu (B).

As I mentioned, context menu is way too crowded. But "Lifecycle" menu can be the place to hold more customized goals (or combinations).
And I don't think it should preset those "clean xxx" by default, but I do agree there should be an approach to configure customized goals in "Lifecycle" menu, or even a separate menu called "Favorites"... 

Similar as pinned tasks in Gradle extension:
image

gustavosimon:
Hi @Eskibear! Sorry about my delayed answer. 

I got it. So, we should create the "clean xxx" goals just in the Lifecycle menu? 

I was thinking about this feature and I think that maybe we can create a extension setting to load the favorites goals into the main menu (context and lifecycle). In this way, we will have this setting with the currently behavior as default. But, if set this setting to true, it will load the favorites goals into the main menu (just to have all goals in the same view). 

What do you think about this solution?
Did you think about another solution to this feature?

Have a nice day! 😃

Eskibear:
>  load the favorites goals into the main menu

I like the idea. But I think a new "Favorites" menu would be more proper. Like, if any command is defined in `maven.terminal.favorites`, we render them "Favorites" menu. Otherwise we don't show the menu.
Or we can always keep the menu to expose this feature, and add a button on the right, as a shortcut to the corresponding setting.

gustavosimon:
I got the idea and like it. 

Can we create a new command to execute the favorite goals from Ctrl + Shift +P (Cmd + Shift + P)? Similar as 'Maven: Execute commands...'. So it will be like 'Maven: Execute favorite commands...'?

Should we create a new issue for develop this feature or we can do it in this PR?

Eskibear:
> Can we create a new command to execute the favorite goals from Ctrl + Shift +P (Cmd + Shift + P)?

There already is one.  https://github.com/microsoft/vscode-maven/blob/2a254ba103708298e8b31172eb89ae4663e72f78/package.json#L229-L233


> Should we create a new issue for develop this feature or we can do it in this PR?

I suggest that you open a new issue, and link it to the discussion. BTW please keep the indentation of package.json unchanged, otherwise we don't easily see what's changed (as in this PR). Thanks.

gustavosimon:
You're right! It's enough for me.
I'm going to create a new issue and work on it. So, can you assign the new issue for me?
Thank you very much!

gustavosimon:
Created a new issue about the discussion on this PR. The new issue is #884. I'm going to close this PR.

PR #877, opened 2022-09-26
can filter dependencies to add by groupId

Eskibear:


Eskibear:
Previously only artifact id can be filtered. 

Now:

image

jdneo (approved):

Issue #884, opened 2022-10-01
Create a new Favorites menu


gustavosimon:
## Suggestion

Create a new Favorites menu to render the custom Favorites commands.
This menu is to improve the experience about developing Java code using this extension.

## Use Cases

I want use this feature to set a custom command and quickly execute the maven goal.
Now, we just execute the favorites/custom commands from Ctrl + Shift + P (Cmd + Shift + P, on Mac)

## Examples

An example is add the 'mvn clean install' command to clean and install the Maven project.
We can also set another commands like 'clean compile', 'clean package', ...
So, in this way we can quickly execute the custom maven goal.

This issue is about the discussion on PR #870.

gustavosimon:
Hi @Eskibear ! Can you assign this issue for me? Thanks.

gustavosimon:
Closing this issue because #900  was merged and #901 was created.

PR #900, opened 2022-11-06
Created new Favorites Menu as described in the issue #884

gustavosimon:


src/explorer/model/FavoriteCommand.ts
```diff
@@ -0,0 +1,23 @@
...
+export class FavoriteCommand implements ITreeItem {
+    
+    constructor(public project: MavenProject, public command: string, public alias: string, public debug?: boolean) {}
+
+    getContextValue(): string {
+        return "maven:favorites";
+    }
+
+    getTreeItem(): TreeItem | Thenable<TreeItem> {
+        const treeItem: vscode.TreeItem = new vscode.TreeItem(this.alias, vscode.TreeItemCollapsibleState.None);
```
Eskibear, line 18: would be better if it displays real 'command' along with 'alias', to distinguish from each other, as below.
```
treeItem.description = this.command;
```

image

Eskibear (changes requested):
Thanks for contribution! Basically it works well. Some nits:
- display `command` in tree item as commented in line.

Further improvements:
a) when `maven.terminal.favorites` is not configured, it's strange to display empty children, and you get no idea whether it's broken or how to add favorite items. 
I would suggest we add a "+" button as inline action of FavoriteMenu, which provides a UI (e.g. quickpick/input) to add corresponding items.

b) Currently when I change the corresponding setting, I have to manually click the refresh button to get the tree items updated. Listening to the change and refresh tree view would be better. In fact it can be implemented by `vscode.workspace.onDidChangeConfiguration` API.

c) Currently the setting is shared among all projects, meaning you set favorite commands for all projects and have no change to customize it for each project. But for each project in the tree view, there is a corresponding menu, which is a little bit inconsistent and strange. Not sure whether we should allow to configure it per project.

For above further thoughts, it's ok not to address in this PR, but to create new issues to track.

gustavosimon:
Hi @Eskibear ! Thanks for feedback 😀!

I will add the `command` in tree item as commented in line. The idea is awesome.

About `a` and `b` items, I really appreciate the idea. I think that we can create a new enhancement issue to improve this behaviors.

About `c` item, I just take the same behavior about the LifecycleMenu, because this commands was list for each project.
How can we track this one?

image

Btw, just for curiousity: What's the meaning of `nits`? It is like a note/observation?

Eskibear:
> About a and b items, I really appreciate the idea. I think that we can create a new enhancement issue to improve this behaviors.

Great!

> About c item, I just take the same behavior about the LifecycleMenu

IMO there's a little difference. Items listed under LifecycleMenu are just basic phases common for every Maven project, even the simplest ones. Favorite commands are customized by users, for example I can add a "spring-boot:run" for a SpringBoot project, but it doesn't make sense to show under a non-SpringBoot project which has no spring-boot-maven-plugin configured.

Anyway, it's good and we can leave it as is now. Let's consider the improvement when we receive requests from community...


> What's the meaning of nits

from "nitpicking", meaning minor things nice to improve. never mind.

gustavosimon:
Nice! I will create the issue tomorrow for we improve this implementation as described on item a and b.

> IMO there's a little difference. Items listed under LifecycleMenu are just basic phases common for every Maven project, even the simplest ones. Favorite commands are customized by users, for example I can add a "spring-boot:run" for a SpringBoot project, but it doesn't make sense to show under a non-SpringBoot project which has no spring-boot-maven-plugin configured.

You're right. It makes sense. Let's consider thinking to improve this behavior to in future.

I saw that the pipelines failed. Should I do something to fix this?

> Anyway, it's good and we can leave it as is now. Let's consider the improvement when we receive requests from community...

Good. Feel free to ask me for help in any issue. I daily use this extension and really would like to help to improve it more.

Eskibear:
linting issue, you can run `npm run tslint` locally to do the check.

```
ERROR: (no-shadowed-variable) /home/runner/work/vscode-maven/vscode-maven/src/Settings.ts[66, 19]: Shadowed name: 'favorites'
ERROR: (no-trailing-whitespace) /home/runner/work/vscode-maven/vscode-maven/src/explorer/model/FavoriteCommand.ts[10, 1]: trailing whitespace
ERROR: (no-trailing-whitespace) /home/runner/work/vscode-maven/vscode-maven/src/explorer/model/FavoriteCommand.ts[23, 1]: trailing whitespace
ERROR: (no-trailing-whitespace) /home/runner/work/vscode-maven/vscode-maven/src/explorer/model/ITreeItem.ts[14, 1]: trailing whitespace
ERROR: (no-trailing-whitespace) /home/runner/work/vscode-maven/vscode-maven/src/utils/Utils.ts[262, 1]: trailing whitespace
```

gustavosimon:
Hello @Eskibear. Fixed linting issues. Can you trigger the pipeline again?

Eskibear:
TS compiler also complains. If you don't mind, I can fix the errors. Then you can take your time to focus on the rest work.

gustavosimon:
Sure! Thank you @Eskibear.

Eskibear (approved):

Eskibear (approved):

Issue #901, opened 2022-11-08
Improvements on Favorites Menu


gustavosimon:
As described on the conversation of pull request #900, this issue is to improve the user experience about the Favorites Menu.

This items was suggestion from @Eskibear:

a) when maven.terminal.favorites is not configured, it's strange to display empty children, and you get no idea whether it's broken or how to add favorite items.
I would suggest we add a "+" button as inline action of FavoriteMenu, which provides a UI (e.g. quickpick/input) to add corresponding items.

b) Currently when I change the corresponding setting, I have to manually click the refresh button to get the tree items updated. Listening to the change and refresh tree view would be better. In fact it can be implemented by 
vscode.workspace.onDidChangeConfiguration API.

gustavosimon:
@Eskibear, can you assign this issue to me?

Eskibear:
@gustavosimon  
We'll ship the feature at the end of Nov. And we want to ship it with the improvement, which would be good for users to get started, especially for those who don't know the feature at all. 

Do you have time to implement above improvement in this week? I'll decide whether to spare some time working on it based on your schedule. Please let me know, thanks!

Eskibear:
add @cscherryy who authored similar features in Gradle extension.

gustavosimon:
Hello @Eskibear ! I'll work on it this week to ship the features. I have already implemented item `b`. So I have to work on item `a`.

gustavosimon:
Can we close this issue @Eskibear ?

PR #906, opened 2022-11-24
Created a UI to add favorites into the user workspace scope.

gustavosimon:
This PR is about #901.


gustavosimon:
Hello @Eskibear! How are you? 

I have just pushed the improvements that we are talking about in improvement issue #901.

I would like to ask if you can push a commit in this PR to add the translation of "Add favorite..." to chinese language into the files package.nls.zh-cn.json and package.nls.zh-tw.json. Unfornately, I really don't know how I should translate it to chinese.

Developing this feature, I was thinking that will be great to have a button to delete a favorite. Do you think that should we implement this in a new improvement issue?

Sorry about my delayed PR. Thank you about your patience.

package.json
```diff
@@ -556,6 +563,16 @@
           "when": "view == mavenProjects && viewItem == maven:dependenciesMenu",
           "group": "inline@0"
         },
+        {
+          "command": "maven.project.addFavorite",
+          "when": "view == javaProjectExplorer && viewItem == maven:favoritesMenu",
```
Eskibear, line 568: I don't think we need this, as javaProjectExplorer is "Java Projects" view from [project manager extension](https://github.com/microsoft/vscode-java-dependency). And we don't contribute menu items there.

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Change Log
 All notable changes to the "vscode-maven" extension will be documented in this file.
 
+## 0.41.0
```
Eskibear, line 4: 0.40.0 is not released yet, you can put it down there

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "vscode-maven",
   "displayName": "Maven for Java",
   "description": "%description%",
-  "version": "0.40.0",
+  "version": "0.41.0",
```
Eskibear, line 5: 0.40.0

src/Settings.ts
```diff
@@ -32,6 +33,12 @@ export namespace Settings {
         workspace.getConfiguration().update("maven.view", "hierarchical", false);
     }
 
+    export function storeFavorite(favorite: FavoriteFormat): void {
+        const favorites: FavoriteFormat[] | undefined = workspace.getConfiguration().get("maven.terminal.favorites");
```
Eskibear, line 37: although default value of `maven.terminal.favorites` is `[]` (defined in package.json), here according to vscode's api, its value is possible to be `undefined`. 

Suggest to change as below
```
const favorites: FavoriteFormat[] = workspace.getConfiguration().get("maven.terminal.favorites") ?? []; 
favorites.push(favorite);
```

src/handlers/favorites/addFavoriteHandler.ts
```diff
@@ -0,0 +1,57 @@
...
+            return undefined;
+        }
+    });
+
+    if (!command) {
+        return;
+    }
+
+    const debugMode = await vscode.window.showQuickPick(
+        ["true", "false"],
```
Eskibear, line 43: Here I perfer to change the order of true and false, making `false` as default, as I think that's the most case. But that's my personal hypothesis only. WDYT?

src/handlers/favorites/addFavoriteHandler.ts
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+import { Settings } from "../../Settings";
+import * as vscode from "vscode";
+
+export async function addFavoriteHandler() {
+
+    const alias = await vscode.window.showInputBox({
```
Eskibear, line 8: Users care the command more then alias, and actually they even don't need an alias, maybe.

So I would prefer to prompt for `command` first, and then `alias`, and use the command as default value when asking to input `alias`. So those who don't care about `alias` can press enter all the way to go.

Eskibear (changes requested):
Thank you! Overall LGTM, with some suggestions for you to consider.

As for the translation, I'll add them myself after you address all the comments.

Eskibear:
> I was thinking that will be great to have a button to delete a favorite. 

a nice-to-have feature, can be implemented in a separate PR later.

src/handlers/favorites/addFavoriteHandler.ts
Eskibear: OR we can be more straight forward, use options:
- Run
- Debug

It can be more user-friendly

gustavosimon:
Thanks about your feedback! I'll develop the suggestions during today.

package.json
```diff
@@ -1,840 +1,841 @@
...
+		"minimatch": "^3.1.2",
+		"semver": "^7.3.8",
+		"vscode-extension-telemetry-wrapper": "^0.13.2",
+		"vscode-languageclient": "^7.0.0",
+		"vscode-languageserver-protocol": "^3.17.2",
+		"vscode-tas-client": "^0.1.63",
+		"which": "^1.3.1",
+		"xml2js": "^0.4.23"
+	},
+	"__metadata": {
```
Eskibear, line 835: the indentation of `package.json ` has been changed, and extra `__metadata` section is added, perhaps formatted by the test library when you run `npm test`. 

Can you remove this section and change the indentation back?

Eskibear: All the other changes LGTM.

Eskibear: Oh forget about this, let me do it, I will also add `zh` translation.

Eskibear (approved):

Issue #932, opened 2023-01-17
The "Resolve Unknown Type" action was not showing up when hovering over the class name.


loralei28:
**Environment**

Operating System: Windows/Mac
JDK version: 17.0.3
Visual Studio Code version: 1.74.3
Java extension version: 1.14.0

**Steps To Reproduce**

1. Open a maven project
2. Open a .java file. In body of a function, type below statements.

```
Gson gson;
StringUtils.capitalize("abc");
```

**Current Result**
Windows
![image](https://user-images.githubusercontent.com/108506566/212857190-5ded072d-8c18-4568-9522-248cc692db63.png)
Mac
![image](https://user-images.githubusercontent.com/108506566/212858840-79b303f4-ec43-45f7-9cbb-3f188afd65e1.png)

**Expected Result**
![image](https://user-images.githubusercontent.com/108506566/212857168-d14cbdcc-fde1-4cdc-9cde-426767335daf.png)

Eskibear:
- Works well with redhat.java@1.3.0
- "No delegate command" error with redhat.java@1.4.0.

Eskibear:
```
org.osgi.framework.BundleException: Could not resolve module: com.microsoft.java.maven.plugin [164]
  Unresolved requirement: Require-Bundle: org.slf4j.api

	at org.eclipse.osgi.container.Module.start(Module.java:463)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel$2.run(ModuleContainer.java:1852)
	at org.eclipse.osgi.internal.framework.EquinoxContainerAdaptor$1$1.execute(EquinoxContainerAdaptor.java:136)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1845)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.incStartLevel(ModuleContainer.java:1786)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.doContainerStartLevel(ModuleContainer.java:1750)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1672)
	at org.eclipse.osgi.container.ModuleContainer$ContainerStartLevel.dispatchEvent(ModuleContainer.java:1)
	at org.eclipse.osgi.framework.eventmgr.EventManager.dispatchEvent(EventManager.java:234)
[...]

PR #933, opened 2023-01-18
fix: add slf4j.api back

Eskibear:
`slf4j.api` is required by `maven-core`. the bundle is no longer required by m2e, thus we add it back manually to work it around.


Eskibear:
fix #932

jdneo (approved):

PR #937, opened 2023-01-21
Create devcontainer.json

ksh5a:
josn23


Eskibear:
Can you explain the purpose adding this file? AFAIK DevContainers like Codespaces will auto pick proper images according to files in the repo.

PR #953, opened 2023-03-13
POM navigation improvements

Eskibear:
Improvements:
- when artifact `version` is not presented directly in the block, list all possible candidates in local maven repo.
- when effective pom is calculated / dependencies tree is calculated, navigate to the corresponding version directly.


Eskibear:
https://user-images.githubusercontent.com/2351748/224646060-b3c71b22-3d4e-4458-a3c4-59b118098af1.mp4

jdneo (approved):

PR #954, opened 2023-03-15
put common goals into submenu

Eskibear:


Eskibear:
image

nickzhums:
is it possible to put a dotted line, or just a line to group the "most common" goals together? for example, i imagine "clean package install" will be way more popular than "test, compile, deploy".

Eskibear:
Technically it's feasible to do that.. For each entry here we register a corresponding command. If you do think that's necessary, then we need a list of those popular commands.

nickzhums:
Can we do this 

image

Eskibear:
yes, of course.

Eskibear:
image

@nickzhums Are you OK with it?

nickzhums (approved):

PR #956, opened 2023-03-16
add profile support

Eskibear:


Eskibear:
https://user-images.githubusercontent.com/2351748/225494284-f37ec98d-ce56-4fb3-82dc-7cabc13b043f.mp4

src/utils/mavenUtils.ts
```diff
@@ -59,6 +60,13 @@ export async function pluginDescription(pluginId: string, pomPath: string): Prom
     return await readFileIfExists(outputPath);
 }
 
+export async function rawProfileList(pomPath: string): Promise<string | undefined> {
+    const outputPath: string = getTempFolder(pomPath);
+    const profileListPath = `${outputPath}.profiles.txt`;
```
jdneo, line 65: Is this a temp file to store the output? Will it be deleted after execution?

src/explorer/model/MavenProfile.ts
```diff
@@ -0,0 +1,39 @@
+import * as vscode from "vscode";
```
jdneo, line 1: license

src/utils/mavenUtils.ts
Eskibear: Yes there is. But it won't be deleted for below reasons:
- In logs, you know what command was executed and which files are generated. You can open it to check the content if error occurs.
- For each project we generate at most one file, and it will be overwritten in future executions. So no much concern about storage.

jdneo: Ok

jdneo (approved):

Eskibear:
We probably can use `-` instead of `!` to construct the command. See https://maven.apache.org/guides/introduction/introduction-to-profiles.html#deactivating-a-profile
A issue created to track it.

> And is this problem why the java project does not update it's source folders to reflect the change in selected profiles? Or is that another issue?

Probably not, @jdneo would know more about source folder related stuffs.

PR #962, opened 2023-04-12
support to create new project without archetype

### LINKED ISSUE #869: Add creating empty maven project without archetypes
## Suggestion
Add an ability to create an empty maven project without archetypes like it can be done in IntelliJ IDEA.
## Use Cases
Quick creation of maven project without removing test sample files that were added by archetype afterwards and editing pom.xml file for convinsing the installed Java version.
## Examples
This behavior can be found in IntelliJ IDEA. You can create an empty maven project with folders structure and the correct Java version will be already specified in pom.xml file.

Eskibear:
fix #869

### DIFF
--- a/src/archetype/ArchetypeModule.ts
+++ b/src/archetype/ArchetypeModule.ts
@@ -15,6 +15,8 @@ import { Utils } from "../utils/Utils";
 import { Archetype } from "./Archetype";
 import { runSteps, selectArchetypeStep, specifyArchetypeVersionStep, specifyArtifactIdStep, specifyGroupIdStep, specifyTargetFolderStep } from "./createProject";
 import { IProjectCreationMetadata, IProjectCreationStep } from "./createProject/types";
+import { promptOnDidProjectCreated } from "./utils";
+
 const REMOTE_ARCHETYPE_CATALOG_URL = "https://repo.maven.apache.org/maven2/archetype-catalog.xml";
 
 export class ArchetypeModule {
@@ -45,7 +47,11 @@ export class ArchetypeModule {
 
         const success: boolean = await runSteps(steps, metadata);
         if (success) {
-            await executeInTerminalHandler(metadata);
+            if (metadata.archetype) {
+                await executeInTerminalHandler(metadata);
+            } else {
+                await createBasicMavenProject(metadata);
+            }
         }
     }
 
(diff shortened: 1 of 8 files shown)


Eskibear:
https://user-images.githubusercontent.com/2351748/231363354-270e0312-feb3-4721-9d70-8f5056670349.mp4

src/archetype/createProject/SelectArchetypeStep.ts
```diff
@@ -62,31 +74,15 @@ export class SelectArchetypeStep implements IProjectCreationStep {
         }
     }
 
-    public async run_simple(metadata: IProjectCreationMetadata): Promise<StepResult> {
```
Eskibear, line 65: @cscherryy I find this method unused. Is it safe to remove?

src/extension.ts
```diff
@@ -280,52 +280,3 @@ async function openPomHandler(node: MavenProject | { uri: string }): Promise<voi
         }
     }
 }
-
```
Eskibear, line 283: moved to `src/archetype/utils.ts` for reuse

src/archetype/createProject/SpecifyTargetFolderStep.ts
```diff
@@ -1,18 +1,44 @@
...
+
         // choose target folder.
-        const result: Uri | undefined = await openDialogForFolder({
-            defaultUri: metadata.targetFolderHint !== undefined ? Uri.file(metadata.targetFolderHint) : undefined,
-            openLabel: "Select Destination Folder"
+        let result: vscode.Uri | undefined = await openDialogForFolder({
+            defaultUri: metadata.targetFolderHint !== undefined ? vscode.Uri.file(metadata.targetFolderHint) : undefined,
+            openLabel: LABEL_CHOOSE_FOLDER
         });
+        while (result && pathExistsSync(path.join(result.fsPath, metadata.artifactId))) {
```
Eskibear, line 28: add prompt for existing folders, asking whether to overwrite

src/archetype/createProject/SelectArchetypeStep.ts
CsCherrYY: looks like the method is introduced by https://github.com/microsoft/vscode-maven/pull/619/files#diff-91272aa6b61f071f70239054717cefd20b164ca21c2a227d03c00195a514814e.

Anyway, IMO we can remove it.

src/archetype/ArchetypeModule.ts
```diff
@@ -140,6 +146,59 @@ async function executeInTerminalHandler(metadata: IProjectCreationMetadata): Pro
...
+        p.report({ message: "Updating pom.xml file...", increment: 20 });
+        const pomUri = vscode.Uri.joinPath(targetUri, "pom.xml");
+        let pomContent = (await workspace.fs.readFile(pomUri)).toString();
+        pomContent = pomContent.replace("${groupId}", groupId);
+        pomContent = pomContent.replace("${artifactId}", artifactId);
+        await workspace.fs.writeFile(pomUri, Buffer.from(pomContent));
+
+        // create source files
+        p.report({ message: "Creating source files...", increment: 20 });
+        await vscode.workspace.fs.createDirectory(vscode.Uri.joinPath(targetUri, "src", "main", "java"));
```
CsCherrYY, line 176: how about saving the whole template project to `./resources/projectTemplate` so that we only need to copy the folder here?

src/archetype/createProject/SpecifyArtifactIdStep.ts
```diff
@@ -13,7 +13,7 @@ export class SpecifyArtifactIdStep implements IProjectCreationStep {
             const inputBox: InputBox = window.createInputBox();
             inputBox.title = "Create Maven Project";
             inputBox.placeholder = "e.g. demo";
-            inputBox.prompt = "Input artifact Id of your project.";
+            inputBox.prompt = "Input artifact Id of your project. (also used as project name)";
```
CsCherrYY, line 16: VS Code will also add a tip with brackets, so the prompt looks like: `Input artifact Id of your project. (also used as project name) (Press 'Enter' to confirm or 'Escape' to cancel...)`, which looks wired for double bracket string.

How about this: `Input artifact Id as well as project name of your project.`

src/archetype/ArchetypeModule.ts
Eskibear: For two reasons:
1. empty folder (e.g. src/test/java) cannot be checked into git tree
2. the folder structure of `src/main/java` is not fixed, depending on `groupId`

src/archetype/createProject/SpecifyArtifactIdStep.ts
Eskibear: I would use `Input artifact Id (also as project name) of your project.` WDYT?

src/archetype/ArchetypeModule.ts
CsCherrYY: ok, let's keep the current implementation

src/archetype/createProject/SpecifyArtifactIdStep.ts
CsCherrYY: sounds good.

CsCherrYY (approved):
LGTM

jdneo (approved):

PR #1031, opened 2024-05-08
use suggested profile syntax

### LINKED ISSUE #995: Fail to disable profile because shell interprets the '!' character
This fails under MacOs on my box because the shell interprets the '!' character before it runs the command. Any way of tweaking this so that Bash and Zsh you could get it to use the Maven alternative syntax -P-disableThisProfile,enableThisProfile ?

_Originally posted by @JFK-DXML in https://github.com/microsoft/vscode-maven/issues/956#issuecomment-1713854639_
            

https://maven.apache.org/guides/introduction/introduction-to-profiles.html#deactivating-a-profile

Maybe we can use `-` instead of `!` to have better compatibility.

imerljak:
Fixes issues with bash/zsh special meaning to ! symbol.

Resolves #995


Eskibear (approved):
lgtm, thanks!
 @testforstephen please take a look and help to verify.

testforstephen (approved):

PR #1056, opened 2024-11-27
enable maven executable path override on workspace level

### LINKED ISSUE #1055: Feature request
## Suggestion

Change the scope of the setting "maven.executable.path" from `machine` to `machine-overridable`?
https://github.com/microsoft/vscode-maven/blob/43618e6dffb6f8a28defa27ea424bd3e07321b7a/package.json#L658-L662

## Use Cases

Since the scope of the setting is `machine` it can only be set in the user settings and cannot be overridden for a specific workspace. When working on multiple projects (workspaces) with different maven versions it would be helpful if the path to the required maven executable could be set individually.

fmeyertoens:
fix #1055

### DIFF
--- a/package.json
+++ b/package.json
@@ -659,7 +659,7 @@
             "type": "string",
             "default": "",
             "description": "%configuration.maven.executable.path%",
-            "scope": "machine"
+            "scope": "machine-overridable"
           },
           "maven.executable.options": {
             "anyOf": [


palminha:
is this project dead?

wenytang-ms (approved):

palminha:
> @fmeyertoens thank you for contributing, you may download the prerelease to have a try! image

@wenytang-ms is there a way to track and correlate that tag in the Version History with github?

PR #1108, opened 2025-11-13
Optional goal parameter for `maven.goal.custom` command

BoykoAlex:


BoykoAlex:
Currently the use case is call it from `spring-tools` vscode extension. The command is executed programatically to execute `process-aot` goal. None of this has been pushed into `main` branch.
Here is the work in progress (but it is nearly completed): https://github.com/spring-projects/spring-tools/pull/1713. The CodeLens would have a command that executes `maven.goal.custom` command with the path to the pom.xml being the first parameter and the `process-aot` goal (or `compile process-aot` as the second parameter. You'd need a sring boot 4 project in the workspace to test it.
Note that the `maven.goal.custom` already has the support for `goal`. The only downside is that it always expects the user to enter the goal via the UI. So, I'm not sure which part of this PR you'd like to test... The project to import into VSCode could be this one from the `spring-tools` repo: https://github.com/spring-projects/spring-tools/tree/aot-metadata/headless-services/spring-boot-language-server/src/test/resources/test-projects/data-repositories-jpa-4
You can run `maven.goal.custom` command via the command palette with the goal `process-aot` to test the "background execution" issue..

chagong (approved):

martinlippert:
@wenytang-ms Tried the pre-release, works perfectly. Would be great to get an official release out with this.

wenytang-ms:
@martinlippert we already ship the 0.45.0 as the latest official release, you may upgrade the maven extension to have a try!

PR #1152, opened 2026-04-21
Improve workspace configuration handling for Maven executable settings

chagong:
Expanded the set of workspace-scoped settings gated by `restrictedConfigurations` to ensure consistent behavior across different workspace trust states.

Changes:
- Added `maven.executable.path`, `maven.executable.preferMavenWrapper`, `maven.terminal.favorites`, and `maven.pomfile.autoUpdateEffectivePOM` to `restrictedConfigurations`
- Added validation for the configured Maven executable path before use in `getMaven()`
- Added workspace trust check before auto-updating effective POM on file change

### DIFF
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -82,7 +82,7 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
     DEFAULT_MAVEN_LIFECYCLES.forEach((goal: string) => {
         registerCommandRequiringTrust(context, `maven.goal.${goal}`, async (node: MavenProject) => executeInTerminal({ command: goal, pomfile: node.pomPath }));
     });
-    registerCommandRequiringTrust(context, "maven.explorer.refresh", async (item: ITreeItem) => {
+    registerCommand(context, "maven.explorer.refresh", async (item: ITreeItem) => {
         item?.refresh ? item.refresh() : MavenExplorerProvider.getInstance().refresh(item);
     });
     registerCommandRequiringTrust(context, "maven.project.effectivePom", async (projectOrUri: Uri | MavenProject) => await Utils.showEffectivePom(projectOrUri));
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -82,7 +82,7 @@ async function doActivate(_operationId: string, context: vscode.ExtensionContext
     DEFAULT_MAVEN_LIFECYCLES.forEach((goal: string) => {
         registerCommandRequiringTrust(context, `maven.goal.${goal}`, async (node: MavenProject) => executeInTerminal({ command: goal, pomfile: node.pomPath }));
     });
-    registerCommand(context, "maven.explorer.refresh", async (item: ITreeItem) => {
+    registerCommandRequiringTrust(context, "maven.explorer.refresh", async (item: ITreeItem) => {
         item?.refresh ? item.refresh() : MavenExplorerProvider.getInstance().refresh(item);
     });
     registerCommandRequiringTrust(context, "maven.project.effectivePom", async (projectOrUri: Uri | MavenProject) => await Utils.showEffectivePom(projectOrUri));
--- a/src/extension.ts
+++ b/src/extension.ts
@@ -197,7 +197,7 @@ function registerPomFileWatcher(context: vscode.ExtensionContext): void {
             contentProvider.invalidate(dependenciesContentUri(project.pomPath));
 
             await project.refresh();
-            if (Settings.Pomfile.autoUpdateEffectivePOM()) {
+            if (vscode.workspace.isTrusted && Settings.Pomfile.autoUpdateEffectivePOM()) {
                 taskExecutor.execute(async () => {
                     await project.refreshEffectivePom();
                     MavenExplorerProvider.getInstance().refresh(project);
--- a/src/utils/mavenUtils.ts
+++ b/src/utils/mavenUtils.ts
@@ -182,11 +182,16 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
     if (mvnPathFromSettings) {
         // expand tilde to deal with ~/path-to-mvn
         const expandedPath: string = expandHome(mvnPathFromSettings);
-        if (await isExecutablePathSafe(expandedPath)) {
+        const safetyResult: "safe" | "use-default" | "abort" = await checkExecutablePathSafety(expandedPath);
+        if (safetyResult === "safe") {
             return expandedPath;
         }
-        // User declined the suspicious path — skip wrapper (also potentially attacker-controlled)
-        // and fall through directly to system-installed Maven
+        if (safetyResult === "abort") {
+            // User chose Open Settings or dismissed — abort the operation
+            return undefined;
+        }
+        // "use-default": User explicitly chose to use default Maven — skip wrapper
+        // (also potentially attacker-controlled) and fall through to system Maven
         mavenOutputChannel.appendLine(`Configured Maven path "${expandedPath}" was declined. Falling back to system Maven.`);
         return await defaultMavenExecutable();
     }
@@ -202,7 +207,7 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
     return await defaultMavenExecutable();
 }
 
-// Set of executable paths the user has explicitly confirmed as safe during this session
+// Set of canonical executable paths the user has explicitly confirmed as safe during this session
 const confirmedExecutablePaths: Set<string> = new Set();
 
 /**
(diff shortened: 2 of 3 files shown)


src/utils/mavenUtils.ts
```diff
@@ -179,9 +179,13 @@ export async function executeInTerminal(options: {
 
 export async function getMaven(pomPath?: string): Promise<string | undefined> {
     const mvnPathFromSettings: string | undefined = Settings.Executable.path(pomPath);
-    if (mvnPathFromSettings) {
+    if (mvnPathFromSettings && vscode.workspace.isTrusted) {
```
wenytang-ms: ## `vscode.workspace.isTrusted` over-blocks user/machine-level config (Blocking)

The guard `mvnPathFromSettings && vscode.workspace.isTrusted` will suppress **all** configured `maven.executable.path` values in untrusted workspaces — including legitimate **user-level (global)** settings.

For example, `maven.archetype.generate` is not trust-gated (`src/extension.ts:92`) and calls `getMaven()` via `ArchetypeModule`. In an untrusted or no-workspace flow, a user's legitimate global `maven.executable.path` would be silently ignored.

**Suggestion:** Use the VS Code `inspect()` API on the configuration to distinguish the source. Only ignore `workspaceValue` / `workspaceFolderValue` when the workspace is untrusted, while still honoring `globalValue`.

src/utils/mavenUtils.ts
```diff
@@ -179,9 +179,13 @@ export async function executeInTerminal(options: {
...
-    if (mvnPathFromSettings) {
+    if (mvnPathFromSettings && vscode.workspace.isTrusted) {
         // expand tilde to deal with ~/path-to-mvn
-        return expandHome(mvnPathFromSettings);
+        const expandedPath: string = expandHome(mvnPathFromSettings);
+        if (await isExecutablePathSafe(expandedPath)) {
+            return expandedPath;
+        }
+        // Unsafe path (relative or inside workspace) — fall through to wrapper/default fallback
     }
```
wenytang-ms, line 197: ## Unsafe path rejection still falls through to repo-local wrapper (Blocking)

When the user declines a suspicious `maven.executable.path` via the prompt, `isExecutablePathSafe` returns `false` and the code falls through to the `preferMavenWrapper` logic. This means a user who explicitly rejected a suspicious binary can still end up executing the repo's `mvnw` — which is equally attacker-controlled.

**Suggestion:** If a configured path is deemed unsafe and the user declines, the function should return `undefined` (abort) rather than continuing to the wrapper/default fallback. At minimum, do not fall through to wrapper resolution.

src/utils/mavenUtils.ts
```diff
@@ -195,6 +199,50 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
...
+        return true;
+    }
+
+    // Relative paths are inherently suspicious — a real Maven installation
+    // always has an absolute path (e.g. /usr/local/bin/mvn, C:\maven\bin\mvn.cmd)
+    if (!path.isAbsolute(executablePath)) {
+        return await promptForExecutableConfirmation(executablePath);
+    }
+
+    // Absolute paths inside a workspace folder are also suspicious
```
wenytang-ms, line 284: ## Path validation lacks canonicalization

`isExecutablePathSafe` checks containment via `vscode.workspace.getWorkspaceFolder()`, which is a good start, but does not handle:
- **Symlinks / junctions** — a path outside the workspace textually could resolve inside it (or vice versa). Consider using `fs.realpath` for canonicalization.
- **Windows case-insensitive paths** — `C:\Repo\mvn` vs `c:\repo\mvn` could bypass or false-positive the check.

**Suggestion:** Canonicalize both the executable path and workspace folder paths before comparison, using `path.resolve` + `fs.realpath` where possible.

src/extension.ts
```diff
@@ -197,7 +197,7 @@ function registerPomFileWatcher(context: vscode.ExtensionContext): void {
             contentProvider.invalidate(dependenciesContentUri(project.pomPath));
 
             await project.refresh();
-            if (Settings.Pomfile.autoUpdateEffectivePOM()) {
+            if (vscode.workspace.isTrusted && Settings.Pomfile.autoUpdateEffectivePOM()) {
```
wenytang-ms, line 200: ## Uncovered code path: `maven.explorer.refresh` → `refreshEffectivePom()`

The `maven.explorer.refresh` command (`src/extension.ts:85`) is registered via `registerCommand` (not `registerCommandRequiringTrust`). While `MavenExplorerProvider.getChildren()` returns `undefined` for untrusted workspaces (hiding the tree), the `PluginsMenu` refresh path can still trigger `refreshEffectivePom()`.

For a security-hardening PR, it would be more robust to close this gap as well — either by trust-gating the refresh command or adding a trust check inside the `refreshEffectivePom` call chain.

src/utils/mavenUtils.ts
```diff
@@ -195,6 +199,50 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
...
+
+    return true;
+}
+
+async function promptForExecutableConfirmation(executablePath: string): Promise<boolean> {
+    const USE_IT = "Allow";
+    const USE_DEFAULT = "Use Default Maven";
+    const choice: string | undefined = await vscode.window.showWarningMessage(
+        `The configured Maven executable path "${executablePath}" points to a location inside the workspace or is a relative path. Using it could be a security risk. Do you want to proceed?`,
+        { modal: true },
```
wenytang-ms, line 302: ## UX improvements for the confirmation dialog

A few suggestions:
- When the user clicks **"Use Default Maven"** (or dismisses the dialog), the current behavior silently falls through to wrapper/default. The user may not realize a *different* Maven is being used. Consider showing a brief notification or logging it.
- Adding an **"Open Settings"** button option would help the user fix the root cause (the suspicious setting) rather than just working around it.
- The `confirmedExecutablePaths` session-level cache design is reasonable, but please add **unit tests** covering: trusted vs untrusted workspace, global vs workspace setting source, relative vs absolute path, and denied path should not fall back to `mvnw`.

src/utils/mavenUtils.ts
```diff
@@ -195,6 +202,92 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
...
+/**
+ * Validates whether a configured Maven executable path is safe to use.
+ * Rejects relative paths and paths that resolve inside a workspace folder,
+ * as these could be attacker-controlled binaries placed in a malicious workspace.
+ * Prompts the user for confirmation and caches their decision for the session.
+ */
+async function isExecutablePathSafe(executablePath: string): Promise<boolean> {
+    if (confirmedExecutablePaths.has(executablePath)) {
+        return true;
+    }
```
wenytang-ms: ## Cache key is not canonicalized

The `confirmedExecutablePaths` Set stores `executablePath` (the pre-`realpath` value) as the key, but `isExecutablePathSafe` now canonicalizes the path via `fse.realpath` further down. If the user clicks **Allow** for a path that resolves through a symlink/junction, and a later call passes a different textual form of the same physical path (e.g. the canonical form, or a differently-cased Windows path), the cache will miss and the user will be prompted again for the same binary.

**Suggestion:** cache the canonicalized path (and look it up with the canonicalized form) so the confirmation is sticky across callers.

src/utils/mavenUtils.ts
```diff
@@ -195,6 +202,92 @@ export async function getMaven(pomPath?: string): Promise<string | undefined> {
...
+    return true;
+}
+
+async function promptForExecutableConfirmation(executablePath: string): Promise<boolean> {
+    const USE_IT = "Allow";
+    const USE_DEFAULT = "Use Default Maven";
+    const OPEN_SETTINGS = "Open Settings";
+    const choice: string | undefined = await vscode.window.showWarningMessage(
+        `The configured Maven executable path "${executablePath}" points to a location inside the workspace or is a relative path. Using it could be a security risk. Do you want to proceed?`,
+        { modal: true },
```
wenytang-ms, line 302: ## Silent fallback on Esc / `Open Settings`  

When the user dismisses the dialog (Esc) or picks **Open Settings**, `promptForExecutableConfirmation` returns `false` and `getMaven()` silently falls through to `defaultMavenExecutable()` — i.e. a Maven command proceeds with a *different* binary than the one the user configured, with only an output-channel log the user is unlikely to see.

Two small improvements would make intent much clearer:
1. When the user chose **Open Settings**, return `undefined` so the current operation aborts rather than silently using system Maven. The user's intent is "let me fix the config first", not "run it anyway with the default".
2. When the user hits Esc (`choice === undefined`), either return `undefined` as well, or surface a brief non-modal `showInformationMessage` so the fallback isn't invisible.

The Allow / Use Default Maven branches are fine as-is.

wenytang-ms (commented):
Two remaining nit-level but user-visible issues from the latest round.

wenytang-ms (approved):

wenytang-ms:
consider to add unit test

wenytang-ms (approved):

chagong:
image

Issue #1041, opened 2024-07-29
maven.terminal.useJavaHome looks at java.home


c365-wreid:
**Describe the bug**
The setting `maven.terminal.useJavaHome` looks at `java.home` which is deprecated and should look for `java.jdt.ls.java.home`

**To Reproduce**
1. Configure `java.jdt.ls.java.home` in a workspace settings file
2. Configure `maven.terminal.useJavaHome` to true in the same settings file
3. use maven command

**Expected behavior**
The maven plugin would check the current VS Code java home setting.

**Environments (please complete the following information as much as possible):**
Version: 1.91.1 (user setup)
Commit: f1e16e1e6214d7c44d078b1f0607b2388f29d729
Date: 2024-07-09T22:06:49.809Z
Electron: 29.4.0
ElectronBuildId: 9728852
Chromium: 122.0.6261.156
Node.js: 20.9.0
V8: 12.2.281.27-electron.0
OS: Windows_NT x64 10.0.22631

**Screenshots**

**Additional context**
Developing Java projects that use installations of Java, Tomcat and Maven relative to my workspace.

cypher256:
Considering why `java.home` was changed to `java.jdt.ls.java.home`, Maven should avoid using `java.jdt.ls.java.home`.
Also, in the near future RedHat's embedded JRE will be version 21, so using `java.jdt.ls.java.home` will be confusing.

Related issue: 
* #495

GreatGreenStar:
I believe the best option to solve this to be a mixture of the above. That is, a variable of its own should exist and be usable if desired BUT if such does not exist ``java.jdt.ls.java.home`` should be checked. Java.home no longer should be checked, as it is deprecated.

wenytang-ms:
@ghost @cypher256 @GreatGreenStar
Fixed in pre-release `0.45.2026051303`.

We chose not to fall back to `java.jdt.ls.java.home` — it controls the JDK that JDT-LS itself runs on (soon to be Red Hat's embedded JRE 21), not the JDK your Maven build should target. Reusing it would mislead users.

Instead, use `maven.terminal.customEnv` to set `JAVA_HOME` for Maven:

```jsonc
"maven.terminal.customEnv": [
  { "environmentVariable": "JAVA_HOME", "value": "/path/to/your/jdk" }
]
```

The deprecated `maven.terminal.useJavaHome` / `java.home` path is removed (#1167); docs updated in #1165.

PR #1165, opened 2026-05-08
docs(jdk): document maven.terminal.customEnv as the JAVA_HOME override for Maven (#1083)

### LINKED ISSUE #1083: JDK selection during Maven build is incorrect.
Hello! Has this problem been solved? I've also encountered the same issue. In the settings.json file of the VSCode project code, the JDK version set is 1.8. However, on the Mac system, the JDK version used is 17. When performing the package operation through the Maven extension, the JDK version 17 is utilized.

Image

Image

wenytang-ms:
Closes #1083
Refs #1041 (cleans up the `java.home` deprecation as a doc-level signal; code removal tracked separately)

## Summary

This PR is a **documentation + diagnostics** fix for the long-standing complaint that `mvn` runs against the wrong JDK. Investigation revealed the capability users wanted already existed (`maven.terminal.customEnv.JAVA_HOME`) but was undiscoverable and silently un-verifiable.

## Background

Issue #1083 collects four user reports of `mvn` picking the system-default JDK rather than the project's target. Across the issue body and 3 follow-up comments, **no user mentioned `java.configuration.runtimes`** — they wanted a Maven-scoped JDK override.

That override already works pre-PR:

- `Settings.getEnvironment` applies `maven.terminal.customEnv` **last**, so a `JAVA_HOME` entry overrides every other source.
- It is wired into all four execution paths: terminal (`mavenTerminal.ts`), spawn (`mavenUtils.ts`), task (`ArchetypeModule.ts`) and debug (`debugHandler.ts`).
- Reload-on-change is already handled by the existing `onDidChangeConfiguration` listener that disposes terminals when `maven.terminal.customEnv` changes.

The user pain was **discoverability**, not behavior. The "customEnv doesn't work" report in #1083 is consistent with editing the setting in an already-open terminal — exactly the kind of "did it actually take effect?" question that is impossible to answer when no output is produced.

## How `JAVA_HOME` is resolved

`Settings.getEnvironment` (the single source of truth for every Maven invocation) builds the env in this order (later writes override earlier ones):

| Priority | Source | Condition | When applied | Status |
|:-:|---|---|---|---|
| 1 (highest) | `maven.terminal.customEnv` `JAVA_HOME` entry | always when set | last — wins over everything below | **recommended** |
| 2 | `java.home` (Red Hat Java) | only if `maven.terminal.useJavaHome: true` | first — overridden by #1 | **deprecated** (this PR) — `java.home` is [deprecated by Red Hat Java](https://github.com/redhat-developer/vscode-java/wiki/JDK-Requirements#setting-the-jdk); the schema now flags `maven.terminal.useJavaHome` itself as deprecated and points users to `maven.terminal.customEnv`. Code removal is tracked in #1041. |
| 3 (lowest) | inherited process / shell env | always | fallback when #1 and #2 are unset | default if nothing is set |

In modern setups, levels #1 and #3 are the only ones that matter in practice. **vscode-maven does not read `java.jdt.ls.java.home` or `java.configuration.runtimes`** — those are JDT-LS settings (literally the JDK that the language server runs on), not Maven settings, and we explicitly do not cross extension boundaries here. Users who want a project-specific JDK for Maven should use `maven.terminal.customEnv`.

After this PR, whichever of #1 or #2 wins emits a single line to the **Maven for Java** output channel, e.g.:

```
Using JAVA_HOME=/Users/me/jdk-21 (source: maven.terminal.customEnv)
```

If neither line appears, you're on #3 (Maven sees whatever `JAVA_HOME` the shell exposes).

> **macOS / Linux caveat for terminal commands.** Levels #1 and #2 are injected into the terminal env *before* the shell starts. If your `~/.zshrc` / `~/.bash_profile` does an unconditional `export JAVA_HOME=…`, the shell profile runs after the injection and will silently overwrite the value. `Troubleshooting.md` now documents this with a `[ -z "$JAVA_HOME" ] && export …` guard pattern. Background commands (effective-pom and other `child_process.spawn` paths) are not affected — they bypass the shell entirely.

## Changes

| File | What |
|---|---|
| `README.md` | Open the `JAVA_HOME` section by default; add a paragraph naming the #1083 scenario (multi-root + "close the terminal first" advisories). |
| `Troubleshooting.md` | Extend the *JAVA_HOME not correctly set* entry with a `maven.terminal.customEnv` JSON recipe, the precedence rules, and the shell-profile-override caveat with a guarded-export workaround. |
| `package.json` + `package.nls*.json` | Mark `maven.terminal.useJavaHome` as deprecated via `deprecationMessage` / `markdownDeprecationMessage`. VS Code's Settings UI will strike it through and tell users to use `maven.terminal.customEnv`. Localized in English, Simplified Chinese, and Traditional Chinese. |
| `src/Settings.ts` | **+5 lines.** Two inline `mavenOutputChannel.appendLine` calls at the two points where vscode-maven decides `JAVA_HOME` — one for the legacy `maven.terminal.useJavaHome + java.home` path, one inside the `customEnv` loop. The log line states the resolved value and the source. Symmetric to the channel's existing diagnostic writes (`Spawn`, `local repository: …`, `Maven path declined fallback`). |
| `test-plans/maven-customenv-javahome.yaml` | New UI autotest that sets `maven.terminal.customEnv.JAVA_HOME=/opt/autotest-fake-jdk-for-customenv`, runs `Maven: Execute Commands... → validate`, and asserts the source tag appears in the **Maven for Java** output channel. |
| `.gitignore` | Ignore `test-results/` produced by the autotest runner. |

## What this PR explicitly does NOT do

In response to review feedback on the first revision:

- ❌ No auto-resolve from `java.configuration.runtimes` — vscode-maven does not read other extensions' settings.
- ❌ No dispose-on-pom-save — the existing `onDidChangeConfiguration` listener on `maven.terminal.customEnv` is sufficient.
- ❌ No new `maven.executable.javaHome` setting — would duplicate `maven.terminal.customEnv.JAVA_HOME`.
- ❌ No code-level removal of `maven.terminal.useJavaHome` + `java.home` — schema deprecation is the user-facing signal; the actual code path stays for backward compatibility and is tracked by #1041 for follow-up removal.

## Validation

- `npm run tslint` — 0 errors (1 pre-existing warning, unrelated)
- `npx tsc -p ./ --noEmit` — clean
- `npm run test:unit` — 21 / 21 passing
- `npm run vscode:prepublish` — webpack OK
- `test-plans/maven-customenv-javahome.yaml` — 7 / 7 steps green

## Diff size

9 files changed, 132 insertions(+), 2 deletions(-) — of which `src/Settings.ts` is **+5**.

### DIFF
--- a/src/Settings.ts
+++ b/src/Settings.ts
@@ -5,6 +5,7 @@ import * as vscode from "vscode";
 import { Uri, workspace } from "vscode";
 import { FavoriteCommand } from "./explorer/model/FavoriteCommand";
 import { MavenProject } from "./explorer/model/MavenProject";
+import { mavenOutputChannel } from "./mavenOutputChannel";
 
 type FavoriteFormat = { alias?: string; command: string; debug?: boolean }
 export class Settings {
@@ -121,6 +122,9 @@ export class Settings {
         if (environmentSettings) {
             environmentSettings.forEach((s: EnvironmentSetting) => {
                 customEnv[s.environmentVariable] = s.value;
+                if (s.environmentVariable === "JAVA_HOME") {
+                    mavenOutputChannel.appendLine(`Using JAVA_HOME=${s.value} (source: maven.terminal.customEnv)`);
+                }
             });
         }
         return customEnv;
@@ -166,6 +170,7 @@ function _getJavaHomeEnvIfAvailable(): { [key: string]: string } {
     const useJavaHome: boolean = Settings.Terminal.useJavaHome();
     const javaHome: string | undefined = Settings.External.javaHome();
     if (useJavaHome && javaHome) {
+        mavenOutputChannel.appendLine(`Using JAVA_HOME=${javaHome} (source: java.home via maven.terminal.useJavaHome)`);
         return { JAVA_HOME: javaHome };
     } else {
         return {};
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ resources/IndexData/index/segments*
 *.jar
 *.tgz
 AGENTS.md
+
+test-results/
--- a/README.md
+++ b/README.md
@@ -80,7 +80,7 @@ Command Palette -> select `Maven: Update Maven Archetype Catalog`.
 
 ## Additional Configurations
 
-<details>
+<details open>
 <summary>JAVA_HOME and Other Environment Variables</summary>
 
 This extension executes Maven by opening a terminal session and then calling Maven in that session.
@@ -100,6 +100,14 @@ Maven requires the JAVA_HOME environment variable to be set. Maven will also loo
     ]
 }
 ```
+
+> **Using a different JDK for Maven than your system default**
+>
+> If `mvn` is picking up the wrong JDK (e.g. the system default JDK 17 is used even though your project targets JDK 1.8), set the `JAVA_HOME` entry in `maven.terminal.customEnv` to the JDK installation you want Maven to use. This setting is honored by every command this extension runs &mdash; both Maven commands launched in the integrated terminal and background commands such as effective-pom generation.
+>
+> The setting is per-folder, so in a multi-root workspace you can configure a different JDK for each folder via the folder's `.vscode/settings.json`.
+>
+> Already-open Maven terminals keep their environment from the moment they were created. After changing this setting, close any existing Maven terminal (or reload the window) so the next command picks up the new value.
 </details>
 
 <details>
--- a/Troubleshooting.md
+++ b/Troubleshooting.md
@@ -35,10 +35,34 @@ Error message can be collected either **directly from the integrated terminal**,
     ```
     In this case, please follow the error message to reset a correct `M2_HOME`.
 
-* `JAVA_HOME` not correctly set.
+* `JAVA_HOME` not correctly set, or the wrong JDK is being used.
     ```
     The JAVA_HOME environment variable is not defined correctly
     This environment variable is needed to run this program
     NB: JAVA_HOME should point to a JDK not a JRE
     ```
     In this case, please specify a correct `JAVA_HOME` environment variable, or re-install JRE/JDK if necessary.
+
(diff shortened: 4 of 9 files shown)


src/Settings.ts
```diff
@@ -160,14 +186,112 @@ function _getMavenSection<T>(section: string, resourceOrFilepath?: Uri | string)
...
+    return fallback?.path;
+}
+
+function _resolveMavenProject(resourceOrFilepath?: Uri | string): MavenProject | undefined {
+    if (!resourceOrFilepath) {
+        return undefined;
+    }
+    const fsPath: string = typeof resourceOrFilepath === "string" ? resourceOrFilepath : resourceOrFilepath.fsPath;
+    return MavenProjectManager.get(fsPath);
+}
```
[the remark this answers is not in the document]
wenytang-ms: implement a more general solution to parse the xml to get the right JDK version

src/extension.ts
```diff
@@ -199,6 +199,10 @@ function registerPomFileWatcher(context: vscode.ExtensionContext): void {
             contentProvider.invalidate(dependenciesContentUri(project.pomPath));
 
             await project.refresh();
+            // The auto-resolved JAVA_HOME may depend on `maven.compiler.*` properties of the
+            // project (see Settings.ts). Dispose existing Maven terminals so the next command
+            // re-evaluates JAVA_HOME against the updated POM.
+            mavenTerminal.dispose();
```
chagong: Do we need to dispose it every time the pom file got changed?

src/extension.ts
```diff
@@ -217,6 +221,7 @@ function registerConfigChangeListener(context: vscode.ExtensionContext): void {
         if (e.affectsConfiguration("maven.terminal.useJavaHome")
             || e.affectsConfiguration("maven.terminal.customEnv")
             || e.affectsConfiguration("java.home") && Settings.Terminal.useJavaHome()
+            || e.affectsConfiguration("java.configuration.runtimes")
```
chagong: I am not sure if we need to do it, getting java runtimes from jdtls settings. I think the original design is to ensure maven extension's behavior only impacted by maven settings, not settings by another extension. We can probably just fix that bug but not necessary to add parse configure from jdtls settings.

wenytang-ms: removed the runtime config

chagong (approved):

Issue #1127, opened 2026-01-15
Noticed IntelliSense stopped working


jonathonmatthewmacleod:
Type: <b>Bug</b>

Used the F1 > Help: Troubleshoot Issue option to work out why my code suggestions through Intellisense weren't working. It identified this extension as the culpret

I noticed that pressing my usual CTRL+SPACE would bring up the modal for intellisense, but it would just say 'Loading...' and never resolve to checking my file hierachary etc 

Extension version: 0.45.1
VS Code version: Code 1.82.2 (abd2f3db4bdb28f9e95536dfa84d8479f1eb312d, 2023-09-14T05:55:25.390Z)
OS version: Windows_NT x64 10.0.26200
Modes:

<details>
<summary>System Info</summary>

|Item|Value|
|---|---|
|CPUs|AMD Ryzen 9 5950X 16-Core Processor             (32 x 3400)|
|GPU Status|2d_canvas: enabled<br>canvas_oop_rasterization: enabled_on<br>direct_rendering_display_compositor: disabled_off_ok<br>gpu_compositing: enabled<br>multiple_raster_threads: enabled_on<br>opengl: enabled_on<br>rasterization: enabled<br>raw_draw: disabled_off_ok<br>video_decode: enabled<br>video_encode: enabled<br>vulkan: disabled_off<br>webgl: enabled<br>webgl2: enabled<br>webgpu: enabled|
|Load (avg)|undefined|
|Memory (System)|63.93GB (38.27GB free)|
|Process Argv|--crash-reporter-id 9eb4aa85-0116-4fc0-bf65-62d3e88c0a09|
|Screen Reader|no|
|VM|0%|
</details><details>
<summary>A/B Experiments</summary>

```
vsliv368:30146709
binariesv615:30325510
nativeloc1:31344060
dwcopilot:31170013
dwoutputs:31242946
copilot_t_ci:31333650
pythonrdcb7:31342333
aj953862:31281341
6abeh943:31336334
cloudbuttont:31379625
3efgi100_wstrepl:31403338
d472g771:31401736
ec5jj548:31422691
cp_cls_c_966_ss:31446431
cp_cls_c_1081:31443829

```

</details>

wenytang-ms:
[...]

PR #1170, opened 2026-05-27
fix(completion): timeout + CancellationToken for pom IntelliSense network calls (#1127)

wenytang-ms:
## What

Defensive hardening of the pom.xml IntelliSense completion path so that an unreachable / slow Maven Central no longer leaves the suggestions popup stuck on `Loading...` forever.

Refs #1127.

## Why

In #1127 a user reports that `Ctrl+Space` brings up the suggestions popup but it stays on `Loading...` indefinitely. We've asked for more info to confirm the root cause (pom.xml vs Java file, network environment, logs), but regardless of the exact root cause in their setup, the existing completion code has a real "infinite Loading…" hazard worth fixing.

Tracing the pom completion provider:

- `src/extension.ts` registers `PomCompletionProvider` for `pom.xml` with trigger characters `.`, `-`, `<`.
- `PomCompletionProvider.provideCompletionItems` previously `await`ed each sub-provider serially and **ignored the `CancellationToken`** VS Code passes in.
- The `ArtifactProvider` fan-out (Maven Central / index / local) was also fully serial: a slow source blocked all the others.
- The HTTPS helper in `src/utils/requestUtils.ts` had:
  - no socket / request timeout — a stuck TCP connection (proxy, DNS black hole, unreachable Maven Central) left the Promise pending forever;
  - no request-level `error` handler, only a response-level one;
  - no awareness of `CancellationToken`.

Combined, that means **one hung HTTPS call to `search.maven.org` is enough to make the completion popup stay on `Loading...` until VS Code is restarted** — matching the symptom in #1127.

## What changed

- `src/utils/requestUtils.ts`
  - Adds a 10 s timeout (`options.timeout` + `req.on('timeout')` → `req.destroy(...)`).
  - Adds a request-level `error` handler.
  - Adds a single `settle` guard so the Promise can only resolve or reject once.
  - Optionally accepts a `vscode.CancellationToken` and aborts the request via `req.destroy()` on cancellation.
  - Introduces a dedicated `CancellationError` class so that completion-side cancellations (which fire on every keystroke as VS Code supersedes the previous token) do **not** spam `console.error`. `getArtifacts` / `getVersions` suppress logging for that class only.
  - Re-checks the `settled` flag immediately after subscribing to `token.onCancellationRequested`, closing a tiny race where a synchronous error from `https.get` could leave the cancellation listener undisposed.
- `src/completion/PomCompletionProvider.ts`
  - Honors the `CancellationToken` (early-out before parsing the doc, and after fan-out).
  - Runs the registered sub-providers concurrently with `Promise.all` + per-provider `catch` so one failing provider can no longer block the others.
- `src/completion/providers/ArtifactProvider.ts`
  - Switches the central / index / local fan-out to `Promise.allSettled`: a network failure on Maven Central still returns local and index results.
  - Forwards the token to `FromCentral` (the only sub-provider that performs real network I/O — `FromIndex` is an in-process JDT.LS RPC and `FromLocal` is a `fast-glob` over `~/.m2/repository`; neither has a public abort API).
- `src/completion/providers/IXmlCompletionProvider.ts` / `IArtifactProvider.ts` / `FromCentral.ts`
  - Optional `token` parameter plumbing. Existing implementations remain compatible.

## Behavior summary

- **Healthy network**: same result set as before, sourced from all three providers. Tiny perf win because the three sources now run concurrently instead of serially.
- **Unreachable Maven Central**: completion popup now resolves within ~10 s with whatever local / index results are available, instead of staying on `Loading...` indefinitely.
- **User cancels (types, moves cursor, presses Esc)**: in-flight HTTPS requests are aborted instead of running to completion.

### Intentional behavior change in `getArtifacts` / `getVersions`

The existing `try/catch` in these helpers only wrapped `JSON.parse`. I widened the `try` to also wrap the `await httpsGet(...)` call. Net effect:

- **Before**: a network-level rejection from `httpsGet` propagated out of `getArtifacts` / `getVersions`. The completion path tolerated this (it caught at a higher level), but the command-handler callers (`src/handlers/addDependencyHandler.ts`, `src/handlers/setDependencyVersionHandler.ts`) would surface an unhandled rejection.
- **After**: any error (timeout, DNS failure, JSON parse failure) returns `[]`. Command handlers now show an empty quick pick instead of failing, which is the safer UX when Maven Central is unreachable.

This is intentional but worth calling out for reviewers.

## Tests

Added `test/unit/requestUtils.test.ts` (9 cases) covering the changed behavior. Tests stub `https` via `proxyquire` and drive the underlying state machine through `data` / `end` / `error` / `timeout` events:

| Case | Asserts |
|---|---|
| short keywords | no HTTPS call, returns `[]` |
| 200 response | parses docs |
| `options.timeout` | equals 10 000 ms |
| pre-cancelled token | short-circuits, no `console.error` |
| cancel mid-flight | `req.destroy` called, no `console.error` |
| non-cancellation error (`ECONNREFUSED`) | returns `[]`, **does** log |
| socket `timeout` event | destroys req with timeout error, logs |
| `getVersions` success | parses docs |
| `getVersions` cancel | silent, returns `[]` |

`.mocharc.json` also gains `"node-option": ["no-experimental-strip-types"]` to work around a Node 22.18+ default that caused newly created `.test.ts` files to be loaded by Mocha through Node's ESM strip-types path instead of the pre-existing `ts-node/register/transpile-only` CommonJS hook (so `require()` was undefined in the new file). With the flag set, all `.test.ts` files load uniformly via ts-node.

## Validation

- `npm run compile` — clean.
- `npm run tslint` — no new warnings.
- `npm run test:unit` — **30 passing** (21 pre-existing + 9 new).
- `npm run webpack` — bundles successfully.

## Notes for reviewer

- `getLatestVersion` and `fetchPluginMetadataXml` also use `httpsGet` and automatically get the new 10 s timeout. I intentionally did not change their public signatures (no caller passes a token today).
- I considered making the 10 s timeout configurable via `maven.completion.timeoutMs` but kept it as a constant for now to minimize surface area; happy to expose it as a setting if reviewers want.
- This change does **not** touch the JDT.LS-side Maven plugin (`jdtls.ext/com.microsoft.java.maven.plugin`). If #1127 turns out to be Java-side IntelliSense rather than pom.xml IntelliSense, the actual root cause likely lives there or in `redhat.java` / JDT.LS, and we'll need the additional info requested on the issue.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -3,6 +3,5 @@
   "spec": "test/unit/**/*.test.ts",
   "extension": ["ts"],
   "timeout": 15000,
-  "reporter": "spec",
-  "node-option": ["no-experimental-strip-types"]
+  "reporter": "spec"
 }
--- a/.mocharc.json
+++ b/.mocharc.json
@@ -3,5 +3,6 @@
   "spec": "test/unit/**/*.test.ts",
   "extension": ["ts"],
   "timeout": 15000,
-  "reporter": "spec"
+  "reporter": "spec",
+  "node-option": ["no-experimental-strip-types"]
 }
--- a/package.json
+++ b/package.json
@@ -825,7 +825,7 @@
     "build-plugin": "node scripts/build-jdtls-ext.js",
     "compile": "tsc -p ./",
     "tslint": "eslint .",
-    "test:unit": "mocha",
+    "test:unit": "node scripts/run-unit-tests.js",
     "test:autotest": "node scripts/run-autotest-plans.js test-plans",
     "watch": "webpack --mode development --watch --progress",
     "webpack": "webpack --mode development",
--- a/scripts/run-unit-tests.js
+++ b/scripts/run-unit-tests.js
@@ -1,0 +1,45 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT license.
+
+/**
+ * Wrapper that invokes mocha for the unit-test suite.
+ *
+ * Why this exists:
+ *   Node 22.6+ ships `--experimental-strip-types` (on by default since 22.18).
+ *   When that flag is enabled, Mocha can load newly added `test/unit/*.test.ts`
+ *   files through Node's native ESM strip-types loader instead of the
+ *   `ts-node/register/transpile-only` CommonJS hook configured in
+ *   `.mocharc.json`. That breaks any test file that uses `require()`
+ *   (e.g. for `proxyquire`) with:
+ *     ReferenceError: require is not defined in ES module scope
+ *
+ *   The fix is to pass `--no-experimental-strip-types` to Node so that all
+ *   `.ts` test files load uniformly via ts-node. But that flag does not
+ *   exist on Node 20, where the project's CI still runs, so we cannot put
+ *   it directly in `.mocharc.json` — Node 20 would reject it as
+ *   `bad option`.
+ *
+ *   This wrapper detects the Node version at runtime and only forwards
+ *   the flag on Node 22.6+. Node 20 runs mocha exactly as before.
+ */
+
+"use strict";
+
+const { spawn } = require("child_process");
+
+const [major, minor] = process.versions.node.split(".").map(Number);
+const supportsStripTypesFlag = major > 22 || (major === 22 && minor >= 6);
+
+const nodeArgs = [];
+if (supportsStripTypesFlag) {
+    nodeArgs.push("--no-experimental-strip-types");
+}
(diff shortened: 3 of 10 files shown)


wenytang-ms:
while in network error, while user don't cancel, there is no popup tooltip; after that, while timeout, the tooltip will show only local part and stop asking for httpget.
before
before1
before2
after
after

chagong (approved):

